PR #7, opened 2021-12-03
Add http headers support for write api.

johncming:
Add support for http headers so that  it can use cortex with the `auth_enabled` option

Signed-off-by: johncming <johncming@yahoo.com>

### DIFF
--- a/pkg/remotewrite/config.go
+++ b/pkg/remotewrite/config.go
@@ -186,12 +186,10 @@ func ParseArg(arg string) (Config, error) {
 		c.KeepNameTag = null.BoolFrom(v)
 	}
 
+	c.Headers = make(map[string]string)
 	if v, ok := params["headers"].(map[string]interface{}); ok {
 		for k, v := range v {
 			if v, ok := v.(string); ok {
-				if c.Headers == nil {
-					c.Headers = make(map[string]string)
-				}
 				c.Headers[k] = v
 			}
 		}
@@ -226,7 +224,7 @@ func GetConsolidatedConfig(jsonRawConf json.RawMessage, env map[string]string, a
 	getEnvMap := func(env map[string]string, prefix string) map[string]string {
 		result := make(map[string]string)
 		for ek, ev := range env {
-			if strings.HasPrefix(ev, prefix) {
+			if strings.HasPrefix(ek, prefix) {
 				k := strings.TrimPrefix(ek, prefix)
 				result[k] = ev
 			}
@@ -286,7 +284,12 @@ func GetConsolidatedConfig(jsonRawConf json.RawMessage, env map[string]string, a
 		}
 	}
 
-	result.Headers = getEnvMap(env, "K6_PROMETHEUS_HEADERS_")
+	envHeaders := getEnvMap(env, "K6_PROMETHEUS_HEADERS_")
+	for k, v := range envHeaders {
+		if _, ok := result.Headers[k]; !ok {
+			result.Headers[k] = v
+		}
+	}
 
 	if arg != "" {
 		argConf, err := ParseArg(arg)
--- a/pkg/remotewrite/config.go
+++ b/pkg/remotewrite/config.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"net/url"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/kubernetes/helm/pkg/strvals"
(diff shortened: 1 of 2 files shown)


pkg/remotewrite/config.go
```diff
@@ -201,6 +223,17 @@ func GetConsolidatedConfig(jsonRawConf json.RawMessage, env map[string]string, a
 		return null.NewBool(false, false), nil
 	}
 
+	getEnvMap := func(env map[string]string, prefix string) map[string]string {
+		result := make(map[string]string)
+		for ek, ev := range env {
+			if strings.HasPrefix(ev, prefix) {
+				k := strings.TrimPrefix(ek, prefix)
```
yorugac, line 230:
```suggestion
		for ek, ev := range env {
			if strings.HasPrefix(ek, prefix) {
				k := strings.TrimPrefix(ek, prefix)
```
I believe this is a typo ^^

pkg/remotewrite/config.go
```diff
@@ -253,6 +286,8 @@ func GetConsolidatedConfig(jsonRawConf json.RawMessage, env map[string]string, a
 		}
 	}
 
+	result.Headers = getEnvMap(env, "K6_PROMETHEUS_HEADERS_")
```
yorugac, line 289: I think it's fine to overwrite values from previous config applications in case of headers but only as addition of values, not full overwrite of the map. So please change here:
```suggestion
	envHeaders := getEnvMap(env, "K6_PROMETHEUS_HEADERS_")
	for k, v := range envHeaders {
		result.Headers[k] = v
	}
```

pkg/remotewrite/config.go
```diff
@@ -175,6 +186,17 @@ func ParseArg(arg string) (Config, error) {
 		c.KeepNameTag = null.BoolFrom(v)
 	}
 
+	if v, ok := params["headers"].(map[string]interface{}); ok {
+		for k, v := range v {
+			if v, ok := v.(string); ok {
+				if c.Headers == nil {
+					c.Headers = make(map[string]string)
+				}
```
yorugac, line 194: Because of the way this function is called this can be simplified down to:
```suggestion
        c.Headers = make(map[string]string)
	if v, ok := params["headers"].(map[string]interface{}); ok {
		for k, v := range v {
			if v, ok := v.(string); ok {			
```

yorugac (changes requested):
Hi @johncming 
I'm sorry for the delay and thank you for your PR! And actually, it's the first contribution to this repo, congrats :tada:

I'll be happy to merge this but first, I have a few requests in the implementation: please see the comments. Sadly, the configs in xk6 extensions are rather tedious right now but I hope we'll be able to improve it soon :sweat_smile: 

And one more thing: could you please add `assert.Equal(t, expected.Headers, actual.Headers)` to the test function here: https://github.com/grafana/xk6-output-prometheus-remote/blob/591f19a7f89a83a1d3009aca91eab013a6c099bb/pkg/remotewrite/config_test.go#L191-L201 and see if the tests continue to work?

johncming:
@yorugac Thank you for the review, it has been fixed.

pkg/remotewrite/config.go
```diff
@@ -253,6 +284,13 @@ func GetConsolidatedConfig(jsonRawConf json.RawMessage, env map[string]string, a
 		}
 	}
 
+	envHeaders := getEnvMap(env, "K6_PROMETHEUS_HEADERS_")
+	for k, v := range envHeaders {
+		if _, ok := result.Headers[k]; !ok {
+			result.Headers[k] = v
+		}
+	}
```
yorugac, line 292: Thank you for the update! There is just one more change needed:

```suggestion
	envHeaders := getEnvMap(env, "K6_PROMETHEUS_HEADERS_")
	for k, v := range envHeaders {
		result.Headers[k] = v
	}
```
In many other cases your logic here would be just right, but this is an official k6 extension and it follows the general guideline of k6 on how config options should behave. The guideline  can be seen here:
https://k6.io/docs/using-k6/options/#using-options

> command-line flags > environment variables > exported script options > config file > defaults

So env var _should_ overwrite the values in JSON. Currently, there's only one test in `GetConsolidatedConfig` for headers and it doesn't catch this but if another test is added, e.g. with the same header name but different values in both JSON and env map, it would fail to adhere to k6 config guideline.

johncming: After fixed, three test cases cover the priority of JSON, env and arg.

yorugac (approved):

yorugac:
@johncming Thank you for all the work and for the additional tests :slightly_smiling_face:

PR #27, opened 2022-06-13
PoC Time series

codebien:
The code is a PoC based on docs/metrics-data-model.md, more research and polish will be applied when a consensus on the design document will be achieved.

It contains a very basic implementation for tdunning's [TDigest](https://github.com/tdunning/t-digest) based on a [3rd-party library](https://github.com/spenczar/tdigest). The math from the library doesn't work so I will replace it soon.


docs/metrics-data-model.md
```diff
@@ -0,0 +1,177 @@
...
+        Value float64
+}
+```
+
+#### Open questions:
+
+* `TimeseriesSample` knows the time series concept and it has the reference to a time series. The Sample is raw data and there isn't yet an identified/matched time series to reference. In this way, the collecting phase doesn't  require knowledge from a time series database. It could speed up the  implementation and introduction of the first stage, where hopefully we don't touch the sample generation (e.g. all the `k6/js/modules`).
+* In the case the two types (`Sample` and `TSample`) are confirmed, we need the `TagSet` in the Sample or can we just simplify it as a `[]string` or `map[string]string` and delegate it in a centralized place the `TagSet` association?
+* Currently, the metric type is defined by the `metrics.Metric` but the same type handle Thresholds and Sink operation. Should the `TimeSeries` type depend on `Metric`? Or should we wait to split them in a different struct?
+
```
codebien, line 53:
```suggestion
* Do we need `Contains ValueType`?
```
It seems used only from the summary. Should we move the mapping directly there? If not, then I think we should extend the [current JS API](https://k6.io/docs/javascript-api/k6-metrics/trend/) for supporting all the possible values (currently, it supports only Time).

docs/metrics-data-model.md
```diff
@@ -0,0 +1,177 @@
...
+* TimeSeries (metric, tagset)
+* Sample (metric, tagset, timestamp, value)
+* TimeSeries Sample (timeseries ref, timestamp, value)
+
+```go
+// This is the current k6 Metric polished from the Thresholds stuff.
+type Metric struct {
+	Name     string     `json:"name"`
+	Type     MetricType `json:"type"`
+	Contains ValueType  `json:"contains"`
```
codebien, line 19:
```suggestion
* Do we need `Contains ValueType`?
```
It seems used only from the summary. Should we move the mapping directly there? If not, then I think we should extend the [current JS API](https://k6.io/docs/javascript-api/k6-metrics/trend/) for supporting all the possible values (currently, it supports only Time).

docs/metrics-data-model.md
```diff
@@ -0,0 +1,177 @@
...
+}
```
codebien, line 20:
```suggestion
Observed bool
}
```
Do we need it with the time series model? Potentially, when the thresholds are parsed we can just initialize a time series and it will be available also if any Sample has been generated.

docs/metrics-data-model.md
```diff
@@ -0,0 +1,177 @@
...
+
+type Sink interface {
+    AddPoint(v float64)
+}
+
+type TSample struct { 
+	TimeSeriesID string
+	Timestamp      uint64
+	Value          float64
+}
```
codebien, line 38:
```suggestion
```
It isn't really required, the engine / ingester will resolve the time series then it invokes `AddPoint(Sample.Value)`. Not really required to have a new struct.

docs/metrics-data-model.md
```diff
@@ -0,0 +1,177 @@
...
+        MetricName string
+        Timestamp uint64
+        Tags TagSet
+        Value float64
+}
+```
+
+#### Open questions:
+
+* `TimeseriesSample` knows the time series concept and it has the reference to a time series. The Sample is raw data and there isn't yet an identified/matched time series to reference. In this way, the collecting phase doesn't  require knowledge from a time series database. It could speed up the  implementation and introduction of the first stage, where hopefully we don't touch the sample generation (e.g. all the `k6/js/modules`).
```
na--, line 50: I think this might not be the best way to go about it :thinking: It would force the calculation of hashes to run in just a single thread. If that was done as soon as a metric is measured in the `js/` package (or elsewhere), that relatively CPU intensive task will be spread apart on multiple cores... :thinking: 

However, if the association with a `TimeSeries` is done there, it will require locking, whereas if it was done centrally from a single thread, it will not require any locking, just a map lookup (or insertion, if the time series didn't exist).

WDYT?

docs/metrics-data-model.md
```diff
@@ -0,0 +1,177 @@
...
+* Should we change something in `metrics.PushIfNotDone()`?
+* Do we require to identify the time series before invoking `metrics.PushIfNotDone`? So, does the sample require to reference the time series?
+
+## metrics.Ingester
+
+TBD
+
+## Outputs
+
+Outputs should interact with the `tsdb` for fetching a time series snapshot. `series := tsdb.GetSeries()`.
```
na--, line 117: Again - some outputs will require more than an aggregated snapshot, they might require access to the raw samples.

docs/metrics-data-model.md
```diff
@@ -0,0 +1,177 @@
...
+
+In the case, we couldn't directly implement the right dynamic histogram, this would help advanced users that know how the SUT behaves to use histogram instead of the 6 gauges workaround.
+
+##### End of series markers
+
+Is the marker required and/or useful for the output?
+
+#### Thresholds
+
+Thresholds should act as an output, it should work on top of a time series data layer. They get the latest value for the time series and check if the value triggers the threshold based on the defined condition.
```
na--, line 156: This is how it already is done after my latest refactoring, see how the `MetricsEngine` receives metric samples by just pretending to be another normal `Output`:
https://github.com/grafana/k6/blob/b725f03be5873b9b0d933b94a83924ff7c1f15e4/metrics/engine/ingester.go#L14-L16
https://github.com/grafana/k6/blob/b725f03be5873b9b0d933b94a83924ff7c1f15e4/core/engine.go#L98-L103

But it can't directly work on the aggregated data from time series since a single threshold spans multiple time series.

docs/metrics-data-model.md
codebien: >Thresholds (and the end-of-test summary) can span multiple time series, see my comment at the top. For example, you might have a threshold on http_req_duration{status:200}, while your actual time series are http_req_duration{status:200,url:http://foo,...},http_req_duration{status:200,url:http://bar,...},http_req_duration{status:200,url:http://baz,...}.

My idea was the summary and the thresholds should use centralized per-time series sinks and aggregate them when required. This should be fine for the summary where we print it just at the end, but in the case of thresholds it will be heavy if we do it for each generated sample.

What is your idea about threshold integration? Should each Threshold keep the Sink autonomously and make the evaluation on top of it?

>We might not need Observed, I don't know, but it doesn't really matter, it will be something local for the MetricsEngine.

Ok, so I'm not including it in this document.

docs/metrics-data-model.md
```diff
@@ -0,0 +1,177 @@
+# k6 Time Series
+
+We would migrate to a time series data model for getting the benefits of aggregated data so just allocating samples one time and pass around the aggregated time series value (or values - e.g. in case of Trend). The Outputs will switch from fetching the buffered samples to fetching a snapshot of the time series (the same concept adopted by Prometheus' scraping operation).
```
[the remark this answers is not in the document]
codebien: ok, I will try to rewrite this part. If we exclude the aggregation from the time series concept what are the other benefits we get?

I see:

* Efficient comparison - we can compare quickly if a time series matches another just by checking the hash.
* Memory storage - we can reference the hash instead to pass the entire name+tags where tags will end in a copy
* Post-run analysis  - We can count the number of generated time series so we know the unique combinations
* Do we have more?
 
This should be also a quick summary of what `1831` is already documenting.

>For example, this output will benefit from both, since Prometheus deals with aggregated data. However, the JSON and CSV outputs certainly wont. The most they will benefit from the grouping of metric samples in time series is to maybe cache some things a bit better. The end-of-test summary will benefit somewhat, but not its aggregation spans multiple time series.

Does it mean we will continue to flush all the samples to the outputs then each output will sink/aggregate autonomously?

go.mod
```diff
@@ -44,6 +44,7 @@ require (
 	github.com/prometheus/client_model v0.2.0 // indirect
 	github.com/prometheus/common/sigv4 v0.1.0 // indirect
 	github.com/prometheus/procfs v0.6.0 // indirect
+	github.com/spenczar/tdigest v2.1.0+incompatible // indirect
```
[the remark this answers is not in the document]
codebien: I already experimented with https://openhistogram.io ([circonusllhist](https://github.com/openhistogram/circonusllhist)). It works better, the Go implementation and the quantile math seem better. The unique cons I see at the moment is the fact we can't get in an efficient way the buckets and relative numbers - we would use the function for byte/string serialization of the entire histogram. In any case, I see it as fixable.

docs/metrics-data-model.md
na--: > What is your idea about threshold integration? Should each Threshold keep the Sink autonomously and make the evaluation on top of it?

Yeah, this is one of the things that I meant with my earlier comment that time series are [orthogonal (i.e. independent, unrelated, etc.)](http://veryfancy.net/blog/orthogonal-concerns/) to aggregation. Determining the "group" ("time series") of a metric `Sample` based on its metric name and tag values is one thing. Aggregating groups of metric samples is quite another, the group won't always be related to the time series.

In the Prometheus outputs (whether remote-write or not), the mapping will probably be 1:1 - you'd have a `Sink` for every time series, and every metric `Sample` will be sent to exactly one `Sink`, right? 

However, in the thresholds and the end-of-test summary, the mapping would be n:m. Each threshold definition can span multiple time-series, since it is defined only by partial (non-exhaustive) constraints. The example I gave with `http_req_duration{status:200}` can match a request to any URL and method, which may all be in different time series. But you can also have another threshold in the same tests that also matches some of them, e.g. `http_req_duration{method:GET}`, and a third threshold `http_req_duration{status:200,method:GET} with both constraints :sweat_smile: So one metric `Sample` will potentially need to be sent to multiple Sinks. And you would not necessarily be able to combine individual per-time-series Sinks back in a single per-threshold sink. In any case, even if you could, it would be quite inefficient to do so every few seconds... :disappointed:  Not to speak about potential future improvements to thresholds which are also going to make things more complicated, e.g. https://github.com/grafana/k6/issues/2379 :sweat_smile: 

That said, even if we can't necessarily reuse or strongly tie together time series and metric sample aggregation/sinks, there  are still some benefits :tada: For example:
- We probably could reuse the same `Sink` implementations here and in other similar outputs and in the `MetricsEngine` (thresholds / end-of-test summary). The grouping will just need to be done differently in both places and decoupled from the aggregation :man_shrugging: In my distributed execution PoC (https://github.com/grafana/k6/pull/2438) I used [circonusllhist](https://github.com/openhistogram/circonusllhist) just for expedience, and we still need to do more research and testing about the various tradeoffs of such algorithms (https://github.com/grafana/k6/issues/763#issuecomment-1059090120), but we probably won't need 2 different ones :man_shrugging: :crossed_fingers: :pray: :sweat_smile: 
- If every metric `Sample` has a link to its time series, [this part of the `MetricsEngine`](https://github.com/grafana/k6/blob/556747fbe585c981507904d634e705c68891e663/metrics/engine/ingester.go#L82-L86) can be implemented much more efficiently :tada: Right now it does nested string comparison for every `SubMetric` definition and on every `Sample`... :disappointed:  With time series, the `MetricsEngine` can create a mapping for every new time series it sees to the matching sub-metrics and reuse that mapping when new samples with the same sub-metric arrive.

go.mod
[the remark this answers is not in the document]
na--: Again, I am not sure this is the best choice, I saw some issues with it during my PoC distributed execution experiments. More R&D is definitely needed... :sweat_smile:

docs/metrics-data-model.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Data model
+
+#### Metrics and Series
+
+* Metric (name, type)
+* TimeSeries (metric, tagset)
+* Sample (metric, tagset, timestamp, value)
```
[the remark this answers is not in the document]
na--: Yeah... :thinking: I'll resolve this thread, we can continue the discussion in https://github.com/grafana/xk6-output-prometheus-remote/pull/27#discussion_r897761049

docs/metrics-data-model.md
```diff
@@ -0,0 +1,177 @@
...
+
+Is the marker required and/or useful for the output?
+
+#### Thresholds
+
+Thresholds should act as an output, it should work on top of a time series data layer. They get the latest value for the time series and check if the value triggers the threshold based on the defined condition.
+
+## Known issues
+
+* Name and URL tags: `k6` is tagging HTTP requests with the URL. It will create a high cardinality issue for the time series data model.
```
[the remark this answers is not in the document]
codebien: @na-- or even better, when we will have the metadata concept then we can just move it there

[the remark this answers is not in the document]
na--: Yes, that's what I'd like, though there will be some... complications :disappointed: The [`iter`, `vu` (and maybe `ip`) tags](https://github.com/grafana/k6/blob/556747fbe585c981507904d634e705c68891e663/metrics/system_tag.go#L103-L106) should probably also be metadata, not tags, but it's a bit too early to see how all of this will work out.

docs/metrics-data-model.md
codebien: I see the central solution as a more straightforward step for introducing this new model in the core. We can extend the solution with the distributed part in the next step.

I collapsed the two structures in the current Sample just adding a TimeSeries pointer that I expect the metrics engine or the ingester will resolve. In this way, we can resolve the submetrics/contains issue in the first step, and in a later PR, we can drop the Tags field from the Sample structure by distributing the time series generation to the components on the edge.

Do you see limits for it?

docs/metrics-data-model.md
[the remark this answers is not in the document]
mstoykov: I have a feeling that we converge on the idea that we don't need time series we just want `SampleTags` to be comparable in `O(1)` time? If you want to call this time series, I am fine with that as well. 

And then we will be aggregating in each output? With the possibility of there being some in core place where we also aggregate them in some manner ... possible in a future version?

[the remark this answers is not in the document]
na--: Kind of... :thinking: If `SampleTags` are comparable in constant time, then they can be used as an index. So that gives us the ability to efficiently build a time series for every unique group of metric tags we want :man_shrugging: But we don't always need the exact same groupings, and we don't always need aggregation. So, by necessity, these things would have to be decoupled.

docs/metrics-data-model.md
na--: > @na-- I see the central solution as a more straightforward step for introducing this new model in the core. We can extend the solution with the distributed part in the next step.

Hmm... maybe? :sweat_smile: isn't the "distributed" part of the time series determination just calculating the hash when the metric sample is assembled? :thinking: That seems like it would be easier to get done at the start, rather than refactoring it later? If it was just dependent on the tags, it could have probably been done by just modifying things [here](https://github.com/grafana/k6/blob/556747fbe585c981507904d634e705c68891e663/metrics/sample.go#L292-L318), but since it also depends on the metric name, it's going to need some more refactoring :disappointed:

docs/metrics-data-model.md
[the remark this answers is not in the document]
codebien: >If SampleTags are comparable in constant time

Just to verify my alignment on this, in the solution in the document we can't really compare `SampleTags` in O(1), we can verify it in O(N). What we can compare in O(1) is the name+SampleTags that in the end most of the time will be the same thing.

But there are exceptions, for example, in the Ingester the first time it sees a new time series it should create a tree of all the parent's time series to know the full set of Sinks to invoke and we have to check all the single tags in the SampleTags for getting the right tree. WDYT? (I have the feeling I'm missing potential limits of this proposal about the future features).

Based on it and that we want distributed sinks, for now, do we really need to involve the name in the hash? In this way, we could avoid to refactor the signature here https://github.com/grafana/k6/blob/556747fbe585c981507904d634e705c68891e663/metrics/engine/ingester.go#L83-L89 mentioned in the other comment.

In the case of sinks per time series, we could have a nested index (e.g `map[metricName]map[SampleTagHash]*Sink`).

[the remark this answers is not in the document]
na--: > Just to verify my alignment on this, in the solution in the document we can't really compare SampleTags in O(1), we can compare it in O(N). What we can compare in O(1) is the name+SampleTags that in the end most of the time will be the same thing.

Yeah, sorry, that's what I meant. If the `metric name + tag set`, i.e. the "time series", is comparable in constant time, we can use it as an index.

> But there are exceptions, for example, [in the Ingester](https://github.com/grafana/k6/blob/556747fbe585c981507904d634e705c68891e663/metrics/engine/ingester.go#L77-L92) the first time it sees a new time series it should create a tree of all the parent's time series to know the full set of Sinks to invoke and we have to check all the single tags in the SampleTags for getting the right tree. Right?

Kind of. When the `MetricsEngine`'s ingester sees a `Sample` with a new (for it) time series, it will need to see which of its own sinks for the `Sample`'s `Metric` match the tags of the new time series. And yeah, that operation is not constant, it's actually `O(m*n)`, where `m` is the number sub-metrics (sub-metric thresholds) and `n` is the length of the constraints for each. However, the improvement here is that this matching `Sink` list (not a tree, just a simple slice) can be cached for the next time a metric from the same time series is encountered! 

> Based on it and that we want distributed sinks, for now, do we really need to involve the name in the hash? In this way, we could avoid refactoring the signature here https://github.com/grafana/k6/blob/556747fbe585c981507904d634e705c68891e663/metrics/sample.go#L292-L318 mentioned in the other comment. WDYT? (I have the feeling I'm missing the potential limits of this proposal regarding the future features).
> 
> In the case of sinks per time series, we could have a nested index (e.g map[metricName]map[SampleTagHash]*Sink)

Hmm, interesting idea, I am not sure which will be better... :thinking: 

On the one hand, if we don't refactor it, we might introduce more complexity everywhere else :thinking: We'd need either these nested maps or composite indexes with `(metric_name, tag_set_hash)` tuples everywhere, right? Even if we ignore potential comparison and indexing inefficiencies, that will add development complexity, I think?

On the other hand, if we don't include the metric name in the hash, we'd be able to potentially more easily correlate Samples from different metrics but with the same tags :thinking: Not sure if that would be useful for anything, since we already package related samples in SampleContainers, but :man_shrugging: 

I think the first (including the metric name in the hash) is the better way to go here, but I am far from sure.

docs/metrics-data-model.md
```diff
@@ -0,0 +1,177 @@
...
+```go
+// This is the current k6 Metric polished from the Thresholds stuff.
+type Metric struct {
+	Name     string     `json:"name"`
+	Type     MetricType `json:"type"`
+	Contains ValueType  `json:"contains"`
+}
+
+type TimeSeries struct {
+    ID uin64 // hash(metric+tags)
```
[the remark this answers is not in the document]
codebien: My assumption: the set of hashes must be unique per test run.

In case of a distributed test (e.g. k6-operator) with the following setup:

Instances: 1000
Unique time series generated from the test: 10k
Total time series = 10M

Where the numbers are nearly unrealistic for our use cases, right?

But even in this case, it seems we are far from the probability for a 50% collision which is around 4 billion ($2^{64/2}$) of generated hashes required. At least this theoretically.

The Go's `hash/maphash` implementation seems not good enough for our case, because it doesn't guarantee the same output across different processes, I don't know if we really require it, but it could be a nice to have for supporting future features. In any case, other solutions like the [xxhash](https://github.com/Cyan4973/xxHash) seem good enough to me.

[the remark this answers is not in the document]
na--: Hmm I've only heard of xxhash, I haven't actually used it, so I can't comment on it too much, but seems like a nice hashing algorithm. But even if we assume it's perfect, I was a bit skeptical about this assertion: 

> But even in this case, it seems we are far from the probability for a 50% collision which is around 4 billion ($2^{64/2}$) of generated hashes required. At least this theoretically.

Because of the ["birthday problem"](https://en.wikipedia.org/wiki/Birthday_problem), collision calculations are somewhat unintuitive. I tried to answer the question "What are the chances that there are collisions with a perfect 64-bit hash in even one out of 100000000 (100m) distinct load tests, each with 100000 (100k) unique time series?", which is probably way beyond even the most realistic worst-case scenario? And I _think_ the I math is somewhat like `1-(e^(-(100000^2)/(2*(2^64))))^100000000`, based on what I remember from statistics and [this article](https://preshing.com/20110504/hash-collision-probabilities/). The result is 2% chance, which is probably good enough for us :sweat_smile: Still, I wonder what hash function and length Prometheus and other similar tools use? :thinking:

codebien:
I simplified the model by removing the TagSet concept, if we are going to use the TimeSeries then it should be enough in terms of memory storage and we can instead benefit from relaxing the system to avoid the synchronization it would introduce.

pkg/remotewrite/remotewrite.go
```diff
@@ -49,116 +53,247 @@ func New(params output.Params) (*Output, error) {
...
 
 		for _, sample := range samples {
-			// Prometheus remote write treats each label array in TimeSeries as the same
-			// for all Samples in those TimeSeries (https://github.com/prometheus/prometheus/blob/03d084f8629477907cab39fc3d314b375eeac010/storage/remote/write_handler.go#L75).
-			// But K6 metrics can have different tags per each Sample so in order not to
-			// lose info in tags or assign tags wrongly, let's store each Sample in a different TimeSeries, for now.
-			// This approach also allows to avoid hard to replicate issues with duplicate timestamps.
-
-			labels, err := tagsToLabels(sample.Tags, o.config)
+			tset, err := o.tagSetFromTags(sample.Tags.CloneTags())
```
na--, line 233: FWIW, this is a bit inefficient, since it will make a copy of the `map[string]string`. Ideally, all of these things should be done with the minimum amount of copying and allocation required. That's only possible on the k6 side, by refactoring the types there directly, so just making a note.

pkg/tsdb/tsdb.go
```diff
@@ -0,0 +1,221 @@
...
+		return (*set)[i].Key > t.Key
+	})
+
+	if len(*set) == index {
+		*set = append(*set, t)
+		return
+	}
+
+	*set = append((*set)[:index], append(make([]*Tag, 1), (*set)[index:]...)...)
+	(*set)[index] = t
```
na--, line 35: Isn't this going to be much less efficient than adding all of the tags unsorted and then sorting the whole slice once at the end? This will just do a binary search on every insertion and then copy (potentially big) chunks of the slice every time an element is added... :confused:

docs/metrics-data-model.md
```diff
@@ -0,0 +1,107 @@
...
+	ID uint64 // hash(metric+tags)
+	MetricName string
+	Tags   []Tag
+}
+
+type Sample struct {
+        TimeSeries *TimeSeries
+        Timestamp uint64
+        Value float64
+}
```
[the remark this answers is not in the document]
codebien: >For one thing, we should probably try to make a lot of these things read-only once they have been created.

A common discussion in the Go world. :smile: I'm not convinced, some questions:

* We will create a not consistent situation between Metric and TimeSeries. For example, why is it relevant to keep TimeSeries protected and to not protect the Metric for a name change? If you are suggesting doing the same for Metric then the impact in the source code seems big.
* How should the Get methods work? I expect they have to make **a copy** of the internal structure. It sounds not optimal in terms of efficiency.

>but we should probably also have some index

I agree.

> so the rest of the codebase could efficiently lookup the value of specific tags?

 Out of curiosity, when do you expect to use it? I expect most of the time `TimeSeries.Tags` access should be used for mapping the **entire** slice because a TimeSeries has been already resolved by matching the ID. In that case, the slice iteration would be ok.

I see the index as useful for the `Contains` operation where we need the pair `key+val` as keys otherwise we need to access by key the map but make a string comparison for the value. Right?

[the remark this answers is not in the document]
na--: > For example, why is it relevant to keep TimeSeries protected and to not protect the Metric for a name change? If you are suggesting doing the same for Metric then the impact in the source code seems big.

Yes, I want to fix `Metric` too. I don't think the impact will be that big, it should be a fairly simple substitution. Most extensions don't use the metric name directly, they use a pointer to the `Metric` object. And besides, we make no promises for backwards compatibility of the Go APIs.

> How should the Get methods work? I expect they have to make a copy of the internal structure. It sounds not optimal in terms of efficiency.

I guess depends on the specific thing. For example, for [`metrics.Metric`](https://github.com/grafana/k6/blob/3e21e678fa13f1cdd442877a105a12f0f255c505/metrics/metric.go#L13-L16), I expect getter methods that make copies of its individual properties will be enough: `m.GetName()` instead of just `m.Name`, etc. Besides, [strings in Go are immutable](https://yourbasic.org/golang/gotcha-strings-are-immutable/), so I don't think it even makes an actual copy, just returns a read-only reference to the original string. 

The problem here is that we [always get a pointer to a `Metric`](https://github.com/grafana/k6/blob/3e21e678fa13f1cdd442877a105a12f0f255c505/metrics/registry.go#L52), we can't have copies of the whole struct since the whole point after the recent refactorings is that we only have a single `Metric` object with the same name. So its properties are mutable :disappointed: 

> 
 Out of curiosity, when do you expect to use it? I expect most of the time `TimeSeries.Tags` access should be used for mapping the **entire** slice because a TimeSeries has been already resolved by matching the ID. In that case, the slice iteration would be ok.
> 
> I see the index as useful for the `Contains` operation where we need the pair `key+val` as keys otherwise we need to access by key the map but make a string comparison for the value. Right?

Yeah, mostly for `Contains()`, since that is needed by the `MetricsEngine` for sub-metric sinks in the end-of-test summary and thresholds. Though that can probably be implemented internally :thinking: Still, I can imagine how some output would like the value of a specific tag (e.g. `url`, `method`, etc.) to do some pre- or post-processing only on specific raw metric samples :man_shrugging: In general though, it should be possible to efficiently inspect a `TimeSeries` and know what are its characteristics.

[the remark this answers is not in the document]
na--: It's fine to have a `GetTags()` function, since sometimes you just need a copy of the tags :man_shrugging: As long as we don't have to make these copies in the normal flow of metric samples, it should be fine. But you shouldn't need to make a copy of the whole tag set in order to get the value for a specific tag.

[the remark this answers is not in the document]
codebien: >since sometimes you just need a copy of the tags

Maybe I'm not considering something but it's not sometimes, it's every time a `flushMetrics` method is invoked for outputs. We need to map the tags and we don't know any specific tag we just have the Sample and the pointer to the TimeSeries. For example, [this line](https://github.com/grafana/k6/blob/3e21e678fa13f1cdd442877a105a12f0f255c505/output/csv/output.go#L224) would be replaced with the `GetTags()` method, right?

docs/metrics-data-model.md
```diff
@@ -0,0 +1,108 @@
...
+
+type Tag struct {
+    Key, Value string
+}
+
+type TimeSeries struct {
+	id uint64 // hash(metric+tags)
+	metricName string
+	tags   []Tag
+	keys map[string]struct{}
```
na--, line 38:
```suggestion
	keys map[string]int
```

or something like that, to allow for direct lookup? :thinking:

docs/metrics-data-model.md
```diff
@@ -0,0 +1,108 @@
...
+The series IDs are `uint64` values generated by hashing the metric name and the tags' key-value pairs.
+
+#### Sink
+
+The sinks are implemented by metric types and they keep the series values up to date:
+
+* Counter:  a monotonic increasing value
+* Gauge: the latest value
+* SparseHistogram: counters per dynamic ranges (buckets) - https://grafana.com/blog/2021/11/03/how-sparse-histograms-can-improve-efficiency-precision-and-mergeability-in-prometheus-tsdb
+
```
na--, line 61: Should we have a separate sink for Rate metrics? In k6 we probably should, but I guess for Prometheus it's just 2 counters?

docs/metrics-data-model.md
```diff
@@ -0,0 +1,108 @@
...
+#### Time Series database (aka tsdb)
+
+We need to store all the time series generated from `k6` run during the execution so the other components (mostly the outputs) could query the storage for getting the value of the time series. It's expected to be in-memory and it should be concurrent-safe.
+
+```go
+type Repository interface {
+    InsertSeries(TimeSeries) error
+    GetSeriesByID(id uint64) (*TimeSeries, error)
+}
```
na--, line 72: Hmm :thinking: You mention that this should be "concurrent-safe", by which I presume you mean that it's safe to call these methods across multiple goroutines? 

If so, then there is a slight problem here, a potential data race more specifically. I see it in the code below as well:
```go
series, err = o.tsdb.GetSeriesByID(id)
if err != nil {
	if !errors.Is(err, tsdb.ErrSeriesNotFound) {
		return 0, err
	}
	series = o.k6MetricToSeries(sample.Metric, tset)
	err = o.tsdb.InsertSeries(series)
	// ...
}
```

Notice that if 2 concurrent goroutines encounter a new time series at the same time, they will both receive a `tsdb.ErrSeriesNotFound` error from `tsdb.GetSeriesByID()`. Then they will both try to call `tsdb.InsertSeries(series)`, but only one of them will succeed.

Instead, I suggest you expose only one method, something like this:
```go
type Repository interface {
    GetOrCreateSeries(id uint64, tags []Tag, metric *Metric) *TimeSeries
    // other methods - GetByID, GetByMetric, GetAll? not sure what else would be required here
}
```

It can do the locking internally and either return the already existing `TimeSeries` or create a new one without any races :man_shrugging: Alternatively, it can probably be the place where hashes are calculated, so you don't need to pass them :thinking: Something like this:
type Repository interface {
    GetOrCreateSeries(tags []Tag, metric *Metric) *TimeSeries // you can query the hash (ID) from the returned *TimeSeries
    // Some other potentially useful methods, though I am not sure if they are needed:
    GetSeriesWithHash(id uint64) (*TimeSeries, error)
    GetSeriesWithMetric(*Metric) []*TimeSeries 
}

docs/metrics-data-model.md
```diff
@@ -0,0 +1,108 @@
...
+* SparseHistogram: counters per dynamic ranges (buckets) - https://grafana.com/blog/2021/11/03/how-sparse-histograms-can-improve-efficiency-precision-and-mergeability-in-prometheus-tsdb
+
+## Storage
+
+#### Time Series database (aka tsdb)
+
+We need to store all the time series generated from `k6` run during the execution so the other components (mostly the outputs) could query the storage for getting the value of the time series. It's expected to be in-memory and it should be concurrent-safe.
+
+```go
+type Repository interface {
```
na--, line 69: This will eventually live in the top [`metrics` k6 package](https://github.com/grafana/k6/tree/master/metrics), right? And it doesn't really store all that much data now that TimeSeries doesn't have built-in sink?

If so, then `Repository` may be a bit misleading, it's more like a `Registry` for time series, it is very similar to the `metrics.Registry` we already have for metrics: https://github.com/grafana/k6/blob/master/metrics/registry.go

So, maybe `TimeSeriesRegistry`? Or just `TSRegistry`? :thinking: Not sure I like these either :sweat_smile:

na--: hmm, or we can just add its few methods to the already existing `metrics.Registry` :thinking: They are very related and we already thread that `Registry` everywhere, and the final "interface" will look fairly tidy and logical, something like this:
```go
type Repository interface {
    NewMetric(name string, typ MetricType, t ...ValueType) (*Metric, error)
    MustNewMetric(name string, typ MetricType, t ...ValueType) *Metric
    Get(name string) *Metric

    GetOrCreateSeries(tags []Tag, metric *Metric) *TimeSeries // you can query the hash (ID) from the returned *TimeSeries
    // Some other potentially useful methods, though I am not sure if they are needed:
    GetSeriesWithHash(id uint64) (*TimeSeries, error)
    GetSeriesWithMetric(*Metric) []*TimeSeries 
}
```

:man_shrugging:

docs/metrics-data-model.md
```diff
@@ -0,0 +1,108 @@
...
+	type     MetricType
+	contains ValueType
+}
+
+type Tag struct {
+    Key, Value string
+}
+
+type TimeSeries struct {
+	id uint64 // hash(metric+tags)
```
na--, line 35:
```suggestion
	id uint32 // sequential id, internal for the 
	hash uint64 // hash(metric+tags)
```

Since we'll have this central thread-safe `Repository` / `Registry`, with locking and everything, we can assign each TimeSeries a unique (in the instance) sequential ID as well, without any overhead :thinking: This might be useful in outputs, since it could potentially allow for faster caches. Instead of a map lookup on every metric, it we could use slices, since these outputs will expect to see metrics with all generated time series anyway. It needs to be benchmarked, since maps are pretty quick as well, but it might help and won't be costly (in either performance or labor) to add :man_shrugging:

docs/metrics-data-model.md
```diff
@@ -0,0 +1,108 @@
...
+```go
+type Repository interface {
+    InsertSeries(TimeSeries) error
+    GetSeriesByID(id uint64) (*TimeSeries, error)
+}
+```
+
+## Samples generation
+
+The current sampling process is controlled by the `metrics.PushIfNotDone` method. All the actual callers should resolve the time series from the storage before push a new Sample, or in the event no one is found to create and insert.
```
na--, line 77: If you add this to the `metrics.Registry` (https://github.com/grafana/xk6-output-prometheus-remote/pull/27/files#r905931538), it should already be propagated everywhere you'll need need this as well, I think :thinking:

docs/metrics-data-model.md
```diff
@@ -0,0 +1,108 @@
...
+It requires the dependency from the time series database for all the callers (e.g. executors, JavaScript modules).
```
na--, line 78: If you add the time series methods to the `metrics.Registry` (https://github.com/grafana/xk6-output-prometheus-remote/pull/27/files#r905931538), it should already be propagated everywhere you'll need need this as well, I think :thinking: So no new dependencies needed :crossed_fingers: .

docs/metrics-data-model.md
codebien: mmm...I thinking if we should do one step more at this point. My idea of tags map was something `map[keySepValue]struct{}` for having the maximum efficiency on `Contains` but it wouldn't support the fetch by `Tag.Key`.

So we could consider this data structure for supporting the single `GetTag(key)`:

```go
tags map[key]stuct{
    value string
    valhash uint64
}
```
In this way, we could support most of the operations with one data structure. The unique missing operation is returning the sorted []Tag for example for `GetTags()`, I see two potential solution:
* store just the keys in a sorted slice
* add another field like `order/index` in the `struct` so we know in which specific place set the `Tag` item.  

```go
func (ts TimeSeries) GetTags() []Tag {
    s := make([]Tag, len(ts.tags))
    for k, data := range {
       s[data.index] = Tag{Key: k, Value: data.Value}
    }
    return s
}
```

The list of operations we have potentially consider in comparisons:

```go
func (TimeSeries) Contains(subset []Tag) bool
func (TimeSeries) HasKey(string) bool 
func (TimeSeries) GetTag(string) Tag
func (TimeSeries) GetTags() []Tag
```

@na-- Thoughts?

docs/metrics-data-model.md
codebien: You're right, I thought about it with the initial reason of having them generated in a central place. Going to update it.

docs/metrics-data-model.md
codebien: >but I guess for Prometheus it's just 2 counters?

I think we could export directly the Rate using a Gauge as we are doing at the moment, but if we want to be sure I guess we should do some better analysis about the expected UX from users.

docs/metrics-data-model.md
codebien: It could be a very good point of optimization. The access to a slice by index should be at least an order of magnitude faster.

I'm just not convinced about the name. If we would have an ID of the TimeSeries I think we should use the `Hash` because it's consistent across instances and runs, it is also not removable from the design because it is an essential concept. So I would prefer to use another name like `SequenceNum{ber}` / `LocalID` / `Index`.

There is a 3rd option where we don't use ID and we use only Hash and another name (e.g `Sequence`).

docs/metrics-data-model.md
codebien: Thinking better on this, it doesn't really work. Because on `Contains` we don't have the relative `tag.valhash` to compare. So I think, we can just support the `GetTag` by binary search :thinking:

na--: I am not sure I understand all of your concerns, or maybe I'm missing something :confused: Please correct me if I'm wrong, but we basically need to be able to efficiently access the tags both in a sorted sequential order and in a random access order (i.e. lookup by their key). So, the basic implementation choices are these:
1. Store the tag keys and values in a sorted list (`[]Tag`) for efficient sequential access and have an index (`map[string]int`) by the key for efficient random access 
2. Store the tag keys and values in a map (`map[string]string`) for the efficient random access and have a sorted list of keys as a `[]string` slice for the sequential access.
3. Something weirder

The first approach seems preferable to me, since the second one involves an extra map per tag access when iterating across all tags sequentially, and anything else seems unnecessary over-engineering at this early point. :man_shrugging: 

Given that these time series objects should be measured in thousands at most, and not millions, I don't see why we'd want anything more complicated than this? You probably won't save all that much memory usage if you don't have the index and use binary search, for example, since I think the strings for the keys will be reused between the slice and the map :man_shrugging: 

Besides, we might be verging in bike-shedding territory here :slightly_smiling_face: . The actual implementation should be internal to the `TimeSeries` object, right? So I don't think it matters very much implementation you pick, we should be able to change it at any time in the future without anything breaking, right?

docs/metrics-data-model.md
na--: Hmm :thinking: How about no ID field at all, to avoid any possible confusion? For example: `hash` and `seqNumberInInstance`?

PR #31, opened 2022-07-13
Send the trend count as a metric

robholland:
Otherwise this information is lost when using prometheus.


robholland:
I think the CLA bot setup is bugged here, it's here twice for some reason.

na--:
Probably a borked retry or something like that :man_shrugging: The commit check status seems fine, so it should be ok  :heavy_check_mark: Thanks for the PR, though I don't have the context to review it, just chimed in about the CLA

codebien (changes requested):
@robholland Thanks for your contribution :bow: 

Can you also update the comment a few lines before? They are now 6 gauges + 1 counter.

codebien (approved):
LGTM, thanks :tada:

Issue #2, opened 2021-10-29
Improve metrics processing and aggregation


yorugac:
The extension must send samples with values that would be understandable to remote agent. To achieve that, k6 aggregation methods of `Sink` interface are used. This leads to the following problems:
1) `Sink`'s `Add` and `Format` methods are not very efficient; additionally, Trend's `Sink` is not fit for such usage at all so ad-hoc optimizations must be used [as is done here](https://github.com/grafana/xk6-output-prometheus-remote/blob/591f19a7f89a83a1d3009aca91eab013a6c099bb/pkg/remotewrite/prometheus.go#L168-L220) (related [k6 issue](https://github.com/grafana/k6/issues/763))
2) attempt to group samples for each metric into `TimeSeries` leads to losing data in tags / labels (related [k6 issue](https://github.com/grafana/k6/issues/1831))
3) converting all samples into separate `TimeSeries` results, at best, in hitting early whatever usage limit is currently set at Prometheus

Since adding any kind of further pre-processing here might clash with the future work on Metric Refactoring at k6, improving these can be considered blocked for now.

Issue #30, opened 2022-07-12
Multiples scenarios show only the aggregated values on prometheus


85humberto:
I'm trying use multiples scenarios and the same metric is sent to all scenarios.

![image](https://user-images.githubusercontent.com/29128791/178594742-aa32b618-4edf-48be-b521-9e5f04ad6cd3.png)
![image](https://user-images.githubusercontent.com/29128791/178594846-ae30313d-3428-44f5-be37-b6a6fb0d756d.png)

In the example above was sent the min value of scenario = cenario4 to all scenarios.
Maybe I'm doing something wrong.

Issue #13, opened 2022-01-20
Bug: incorrect grouping of metrics during aggregation


yorugac:
Described in community forum: https://community.k6.io/t/counters-not-starting-from-zero/2737
Both actual and expected behavior are included in the above description.

The cause: internal aggregation of metrics in the extension is happening by names of metrics and doesn't take labels into account. Ideally, this should be resolved by https://github.com/grafana/k6/issues/1831. However, that issue may take quite a lot of time to resolve fully, so it'd be good if a quicker solution can be added here first. E.g. once some design solutions proposed to https://github.com/grafana/k6/issues/1831, it might make sense to evaluate them here, in the extension first.

Currently considered blocked by https://github.com/grafana/k6/issues/1831 and should be re-visited upon progress there.

mschenk42:
We've noticed trend metrics for requests with labels (name label in particular) are not correct.  When grouping by the name label (request) the trends are all the same for each name label.  Would this issue be why that is the case?

codebien:
Hi @mschenk42, yes, it is the same issue. Hopefully, we will get fixed in the next releases, we are actively working on the main issue.

In the meanwhile, you could consider to workaround creating [custom metrics](https://k6.io/docs/using-k6/metrics/#custom-metrics) per endpoint, you probably need to tune the [System Tags](https://k6.io/docs/using-k6/tags-and-groups/#system-tags) so `k6` doesn't generate too much time series.

Issue #11, opened 2021-12-23
Occasional "duplicate sample for timestamp" errors


Vlad-Stryapko:
I use the latest version of the extension with 0.35.0 k6. Occasionally during the test I get the following error:

> ERRO[0370] Failed to store timeseries.                   error="server returned HTTP status 400 Bad Request: duplicate sample for timestamp"

The errors are not terribly frequent, but not rare either - I've got about 25 errors after 10 minutes of the test run with 255 VUs generating the load. 

Is it possible to somehow mitigate the issue so that the samples are not dropped? Also, is it possible to asses the number of duplicate samples, i.e. how many of them were not written because of the error?

yorugac:
Hi @Vlad-Stryapko
Thanks for opening the issue!

Currently I'd say it isn't about "mitigating this issue" but figuring out when it happens in the first place. I've briefly looked at this issue with duplicate timestamps a few months ago but it didn't seem to behave as an easily repeatable bug (at least back then). Did you happen to notice any specific pattern when these errors happens? Maybe it's a specific k6 script or metric / tag or something in Prometheus setup, etc.

We might end up adding hacks in extension as a workaround for this error like you suggest but first, I think it makes sense to try to gather more info about it and additionally, to try out this recent improvement on Prometheus side https://github.com/prometheus/prometheus/pull/9894 and see if it helps to debug the origin of the issue.

As for the amount of such errors, I don't think Prometheus has separate metric for duplicate timestamps in remote write :disappointed: It does have metric `prometheus_target_scrapes_sample_duplicate_timestamp_total` but by its name, it's probably specific for scraping.

Vlad-Stryapko:
I don't think there is anything specific I've noticed. My test algorithm is pretty straightforward: I send a message to a websocket channel and when I receive a response, I add the elapsed time between a request and response to the trend. 

What I think I can do is as follows: try to repro the bug on my local machine with the latest Prometheus installed; try to reproduce it with a very basic k6 test which would simply write to a trend as quickly as possible. 

As to finding out the dropped data, in scope of the repro I can also introduce a counter of how many times the `trend.Add` was called and then compare the counter to the number of samples in Prometheus. I would assume that generally the % lost should be the same for the same configuration, e.g. 10% of the samples dropped.

Vlad-Stryapko:
I was able to repro the issue with the latest Prom and k6 locally. 

[...]

PR #38, opened 2022-08-31
Sink by time series

codebien:
A simple attempt of aggregating by time series, it fixes the two main issues regarding:

1. Values are now aggregated by time series and not only by the metric name
2. If the containers contain multiple samples in a sub-milliseconds timeframe, theirs values will be aggregated in a single time series using the most recent timestamp

It acts as an intermediate step for the adoption of the new k6 time series model because it starts to use logic by time series so the migration should be simplified.

I will add more tests if we will agree on the approach, potentially adding only tests that can be easily migrated to the new model.

Closes https://github.com/grafana/xk6-output-prometheus-remote/issues/2 https://github.com/grafana/xk6-output-prometheus-remote/issues/30 https://github.com/grafana/xk6-output-prometheus-remote/issues/13 https://github.com/grafana/xk6-output-prometheus-remote/issues/11


yorugac (approved):
I haven't grok-ed the new time series yet but LGTM :star_struck: Ivan, thank you for implementing this!

Out of curiosity, was there a decision to support only Prometheus remote write and no other provider? Since you removed `Mapping`.

Also, just a suggestion: I'd add a comment on timestamp precision. If truncation is the norm during sample conversion, it'd be better to be explicit about it.

codebien:
>Out of curiosity, was there a decision to support only Prometheus remote write and no other provider? Since you removed Mapping.

@yorugac It was an attempt for simplifying the API considering that in the end we are not really using this abstraction. But I feel I lack of knowledge on this point. During the original development did you find implementations where the current mapping would be an issue?

>Also, just a suggestion: I'd add a comment on timestamp precision. If truncation is the norm during sample conversion, it'd be better to be explicit about it.

:+1:

yorugac:
@codebien remote write is supported by [different providers](https://prometheus.io/docs/operating/integrations/) and IIRC, not all of them are using the same formatting or types of metrics as Prometheus. E.g. Prometheus relies on suffixes only for histograms (`_bucket`, etc.) but one of those integrations is relying on suffixes for counters. The remote write protocol doesn't impose rules on time series format but provider (receiver) might have such rules. So the question is basically whether k6 intends to support Prometheus remote write only or remote write in general. That's how `Mapping` came about - assuming the second case.

Maybe it doesn't matter much now? But it's at least something to be aware of then.

codebien:
>Maybe it doesn't matter much now? But it's at least something to be aware of then.

Yeah, I think so, we should add a test suite for checking all the supported remote write implementations and if we will need then we will abstract better the code. I will open an issue for it. It would be also useful to mention them in the README.

pkg/remotewrite/remotewrite.go
```diff
@@ -94,71 +93,183 @@ func (o *Output) flush() {
...
+	var newts []prompb.TimeSeries
+
+	mapMonoSeries := func(s TimeSeries, t time.Time) prompb.TimeSeries {
+		return prompb.TimeSeries{
+			Labels: append(MapTagSet(swm.Tags), prompb.Label{
+				Name:  "__name__",
+				Value: fmt.Sprintf("%s%s", defaultMetricPrefix, swm.Metric.Name),
+			}),
+			Samples: []prompb.Sample{
+				{Timestamp: timestamp.FromTime(t)},
```
yorugac, line 194:
```suggestion
			Samples: []prompb.Sample{
			        // here timestamp is rounded down from ns to ms precision
				{Timestamp: timestamp.FromTime(t)},
```
I meant something like this. Sorry for not being more clear yesterday :sweat_smile:

yorugac (approved):

pkg/remotewrite/remotewrite.go
codebien: This comment plus more tests have triggered good thoughts on my side :+1:  and I caught some issues so I added a better optimization where we use directly the truncated time otherwise some cases wouldn't be covered as described in the comments.
Now the truncate operation is explicit in the logic so we shouldn't need one more comment.'

codebien:
@yorugac I added more steps so please review again. I'm merging this PR now but if you find more issues, I will add them in a second PR. :pray:

pkg/remotewrite/remotewrite.go
yorugac: Glad to hear that!

:+1: I'll re-read the PR on Mon and let you know if I notice anything else.

pkg/remotewrite/remotewrite.go
```diff
@@ -89,76 +88,225 @@ func (o *Output) flush() {
...
+				}
+
+				// If current == previous:
+				// the current received time before being truncated had a higher precision.
+				// It's fine to aggregate them but we avoid to add to the seen map because:
+				// - in the case it is a new flush operation then we avoid delivering
+				//   for not generating duplicates
+				// - in the case it is in the same operation but across sample containers
+				//   then the time series should be already on the seen map and we can skip
+				//   to re-add it.
```
yorugac, line 181: @codebien When does this case actually happen? The same metric, out of order (i.e. with `ts2 <= ts1`) and in different sample containers?

I wonder if it shouldn't be a warning / error level as well.

PR #43, opened 2022-09-27
Refresh the CI steps

codebien:


.github/workflows/all.yml
```diff
@@ -14,43 +14,20 @@ defaults:
...
           fetch-depth: 0
       - name: Install Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v3
         with:
-          go-version: 1.17.x
+          go-version: 1.18.x
       - name: Install golangci-lint
         working-directory: /tmp
         run: go install "github.com/golangci/golangci-lint/cmd/golangci-lint@v$GOLANG_CI_VERSION"
```
olegbespalov, line 33: It's not critical, but if we use the action, like:
```
- name: golangci-lint on Go
        uses: golangci/golangci-lint-action@v3
        with:
          version: ${{ $GOLANG_CI_VERSION }}
```
It benefits us by commenting in PRs, but it's not a big deal

olegbespalov (approved):

Issue #47, opened 2022-10-13
`K6_KEEP_URL_TAG` option not working in `v0.0.4`


BlakeSearlSonocent:
Hi, thank you for this excellent extension.

We've recently tried upgrading to `v0.0.4` as there was a bug fix that we need in k6 `v0.40.0`.

We were using the `K6_KEEP_URL_TAG` option, set to `false` as we have a fair few `url`s that incorporate a uuid. This was working in `v0.0.3`.

Since the upgrade, we have found that the `url` tag is arriving in prometheus again, despite leaving `K6_KEEP_URL_TAG="false"` set. We've also tried `K6_KEEP_URL_TAG=false` as I noticed that the option is expecting a `Bool` but no luck there either.

Would you be able to offer any help? Thanks!

BlakeSearlSonocent:
It looks like the support for it was dropped in this commit: https://github.com/grafana/xk6-output-prometheus-remote/commit/253b51ec74c17d20b6a6da82d5517e7f12a0c7e1

Here is the PR that added the option: https://github.com/grafana/xk6-output-prometheus-remote/pull/17

codebien:
Hi @BlakeSearlSonocent,
it was removed mostly to make this extension more k6 native. Typically this is handled in `k6` using [URL-grouping](https://k6.io/docs/using-k6/http-requests/#url-grouping) or controlling the reported [system tags](https://k6.io/docs/using-k6/k6-options/reference#system-tags).

Does it not work for your use case? If not, can you expand, please?

BlakeSearlSonocent:
Hey @codebien,

That makes a lot of sense, thank you for explaining and providing the resources. We're were using `URL-grouping` but not `system tags`. Dropping `url` from system tags works perfectly for us.

Thanks for your help 😃

Issue #5, opened 2021-10-29
Filtering metrics by tags


yorugac:
To avoid quickly overshooting server side limits, it would be good to have configurable option on how many metrics and how many labels are sent via configuration and perhaps, which ones are sent in the first place.

Partially related [k6 issue](https://github.com/grafana/k6/issues/1321)

yorugac:
I'd say this feature is needed for outputs in general and probably should be part of the core :sweat_smile: 

This issue is not about system tags but about more flexible configuration in general, including custom tags. E.g. Prometheus allows quite a [versatile configuration](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config) for which metrics and labels to keep and which to discard. But that's configuration for scraping, remote write endpoint doesn't have that, AFAIK. Plus, the issue of slow metrics processing from k6 side has been imposing limits on how much can be processed so filtering seemed like a good addition. That's how this issue appeared initially.

Now that I look at it, it can be read as a duplicate for https://github.com/grafana/k6/issues/1321. If so, then it makes sense to close this one: this feature request is more fitting to k6 core anyway. Just to be clear, I consider this a valid feature request regardless of metrics refactoring: as a user, I don't always want to have all metrics stored and system tags option is insufficient.

> I think we should remove [these options](https://github.com/grafana/xk6-output-prometheus-remote/blob/6d82624bb13c70181ec93c69b3827ab15f751475/pkg/remotewrite/config.go#L40-L41) from the configuration.

For sure. These were really just temporary options until something better appears.

yorugac:
> remote write endpoint doesn't have that, AFAIK

Correction: remote write endpoint can probably rely on [`metric_relabel_configs` option](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs) to filter metrics received from k6. But that'd be on Prometheus' side, so *after* metrics were already sent by k6 but before they are stored. And this option may not be present for other remote write outputs. Still, better than nothing.

smazurov:
why is `systemTags` insufficient for basic pairing down of whats sent?

codebien:
Currently, we're not planning to implement this. As we mentioned, we want to address it with a wider approach. If required, use one of the collector available out there to filter out metrics or use one of the previously mentioned workarounds.

PR #48, opened 2022-10-21
Polish config

codebien:
This PR includes changes for the config and the URL printing in the description.

Renamed for consistency the following env var options:

`K6_PROMETHEUS_FLUSH_PERIOD` => `K6_PROMETHEUS_PUSH_INTERVAL` (it aligns to the same name used by timescaledb, influxdb and statsd outputs).

`K6_PROMETHEUS_USER` => `K6_PROMETHEUS_USERNAME` (it aligns to the same name used by influxdb output).

Removed completely for aligning to what we are doing with other outputs the following options:

`KeepTags`, `KeepNameTag`, `KeepUrlTag`, as explained in https://github.com/grafana/xk6-output-prometheus-remote/issues/47, tags should be controlled by `systemTags` option or by the solution explained in https://github.com/grafana/xk6-output-prometheus-remote/issues/5 (TLDR; https://github.com/grafana/k6/issues/1321) when we will have implemented it.

Removed `Mapping` option because at the moment the output does not support different kinds of mappings so it does not make sense to export an option.

[The documentation](https://k6.io/docs/results-output/real-time/prometheus/#options) must be aligned when we will have merged this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,11 @@ K6_PROMETHEUS_REMOTE_URL=http://localhost:9090/api/v1/write ./k6 run script.js -
 
 Add TLS and HTTP basic authentication:
 ```
-K6_PROMETHEUS_REMOTE_URL=https://localhost:9090/api/v1/write K6_PROMETHEUS_INSECURE_SKIP_TLS_VERIFY=false K6_CA_CERT_FILE=example/tls.crt K6_PROMETHEUS_USER=foo K6_PROMETHEUS_PASSWORD=bar ./k6 run script.js -o output-prometheus-remote
+K6_PROMETHEUS_REMOTE_URL=https://localhost:9090/api/v1/write \
+K6_PROMETHEUS_INSECURE_SKIP_TLS_VERIFY=false \
+K6_PROMETHEUS_USERNAME=foo \
+K6_PROMETHEUS_PASSWORD=bar \
+./k6 run script.js -o output-prometheus-remote
 ```
 
 Note: Prometheus remote client relies on a snappy library for serialization which could panic on [encode operations](https://github.com/golang/snappy/blob/544b4180ac705b7605231d4a4550a1acb22a19fe/encode.go#L22).
--- a/pkg/remotewrite/config.go
+++ b/pkg/remotewrite/config.go
@@ -16,43 +16,31 @@ import (
 )
 
 const (
+	defaultURL               = "http://localhost:9090/api/v1/write"
 	defaultPrometheusTimeout = time.Minute
-	defaultFlushPeriod       = 5 * time.Second
+	defaultPushInterval      = 5 * time.Second
 	defaultMetricPrefix      = "k6_"
 )
 
 type Config struct {
-	Mapping null.String `json:"mapping" envconfig:"K6_PROMETHEUS_MAPPING"`
-
-	Url null.String `json:"url" envconfig:"K6_PROMETHEUS_REMOTE_URL"` // here, in the name of env variable, we assume that we won't need to distinguish between remote write URL vs remote read URL
-
+	// here, in the name of env variable, we assume that we won't need to distinguish between remote write URL vs remote read URL
+	URL     null.String       `json:"url" envconfig:"K6_PROMETHEUS_REMOTE_URL"`
 	Headers map[string]string `json:"headers" envconfig:"K6_PROMETHEUS_HEADERS"`
 
 	InsecureSkipTLSVerify null.Bool   `json:"insecureSkipTLSVerify" envconfig:"K6_PROMETHEUS_INSECURE_SKIP_TLS_VERIFY"`
-	CACert                null.String `json:"caCertFile" envconfig:"K6_CA_CERT_FILE"`
+	Username              null.String `json:"username" envconfig:"K6_PROMETHEUS_USERNAME"`
+	Password              null.String `json:"password" envconfig:"K6_PROMETHEUS_PASSWORD"`
 
-	User     null.String `json:"user" envconfig:"K6_PROMETHEUS_USER"`
-	Password null.String `json:"password" envconfig:"K6_PROMETHEUS_PASSWORD"`
-
-	FlushPeriod types.NullDuration `json:"flushPeriod" envconfig:"K6_PROMETHEUS_FLUSH_PERIOD"`
-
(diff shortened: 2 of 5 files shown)


pkg/remotewrite/config.go
```diff
@@ -16,43 +16,33 @@ import (
...
-	Url null.String `json:"url" envconfig:"K6_PROMETHEUS_REMOTE_URL"` // here, in the name of env variable, we assume that we won't need to distinguish between remote write URL vs remote read URL
-
+	// here, in the name of env variable, we assume that we won't need to distinguish between remote write URL vs remote read URL
+	URL     null.String       `json:"url" envconfig:"K6_PROMETHEUS_REMOTE_URL"`
 	Headers map[string]string `json:"headers" envconfig:"K6_PROMETHEUS_HEADERS"`
 
 	InsecureSkipTLSVerify null.Bool   `json:"insecureSkipTLSVerify" envconfig:"K6_PROMETHEUS_INSECURE_SKIP_TLS_VERIFY"`
-	CACert                null.String `json:"caCertFile" envconfig:"K6_CA_CERT_FILE"`
-
-	User     null.String `json:"user" envconfig:"K6_PROMETHEUS_USER"`
```
olegbespalov, line 34: We should also update the README.md

pkg/remotewrite/config.go
```diff
@@ -243,18 +201,14 @@ func GetConsolidatedConfig(jsonRawConf json.RawMessage, env map[string]string, a
 	}
 
 	// envconfig is not processing some undefined vars (at least duration) so apply them manually
-	if flushPeriod, flushPeriodDefined := env["K6_PROMETHEUS_FLUSH_PERIOD"]; flushPeriodDefined {
-		if err := result.FlushPeriod.UnmarshalText([]byte(flushPeriod)); err != nil {
+	if flushPeriod, flushPeriodDefined := env["K6_PROMETHEUS_PUSH_INTERVAL"]; flushPeriodDefined {
```
olegbespalov, line 204: Let's also change the name of variable (migrate to `pushInterval*`)

pkg/remotewrite/remotewrite_test.go
```diff
@@ -9,8 +9,19 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.k6.io/k6/metrics"
+	"gopkg.in/guregu/null.v3"
 )
 
+func TestOutputDescription(t *testing.T) {
+	o := Output{
```
olegbespalov, line 16:
```suggestion
	t.Parallel()
	o := Output{
```

olegbespalov (commented):
Generaly looks good, I left a few comments :+1:

pkg/remotewrite/config.go
```diff
@@ -243,18 +201,14 @@ func GetConsolidatedConfig(jsonRawConf json.RawMessage, env map[string]string, a
 	}
 
 	// envconfig is not processing some undefined vars (at least duration) so apply them manually
-	if flushPeriod, flushPeriodDefined := env["K6_PROMETHEUS_FLUSH_PERIOD"]; flushPeriodDefined {
-		if err := result.FlushPeriod.UnmarshalText([]byte(flushPeriod)); err != nil {
+	if flushInterval, flushIntervalDefined := env["K6_PROMETHEUS_PUSH_INTERVAL"]; flushIntervalDefined {
```
olegbespalov, line 204: It seems like the varial still has a `flush` in the name, but it's not super-critical

olegbespalov (approved):

pkg/remotewrite/config.go
codebien: Yeah, I confused the two names :facepalm:

README.md
```diff
@@ -25,7 +25,11 @@ K6_PROMETHEUS_REMOTE_URL=http://localhost:9090/api/v1/write ./k6 run script.js -
 
 Add TLS and HTTP basic authentication:
 ```
-K6_PROMETHEUS_REMOTE_URL=https://localhost:9090/api/v1/write K6_PROMETHEUS_INSECURE_SKIP_TLS_VERIFY=false K6_CA_CERT_FILE=example/tls.crt K6_PROMETHEUS_USER=foo K6_PROMETHEUS_PASSWORD=bar ./k6 run script.js -o output-prometheus-remote
+K6_PROMETHEUS_REMOTE_URL=https://localhost:9090/api/v1/write \
+K6_PROMETHEUS_INSECURE_SKIP_TLS_VERIFY=false \
+K6_PROMETHEUS_CA_CERT_FILE=example/tls.crt \
```
codebien, line 30: Do we need this? The user could use the Go env vars `SSL_CERT_FILE` and `SSL_CERT_DIR` for passing the location or the dir of the root store.

olegbespalov: It's a good question :thinking: 

Honestly, I think that we can stick with the default golang's environment variable if that works precisely the same way, but I guess it's still important to mention that because maybe not all of our users are golang devs. But I'm not so strict in that opinion.

codebien: @yorugac WDYT?

yorugac: Apologies for the delay! This message totally got lost in my Github subscriptions.
This env var appeared in order to connect to Prometheus set up with [auth](https://prometheus.io/docs/prometheus/latest/configuration/https/): if there is another way to connect to such an instance without using env var, it should be checked. 

Btw, Mimir has something similar in its setup to Prometheus: https://grafana.com/docs/mimir/latest/operators-guide/secure/securing-communications-with-tls/#configure-tls-certificates-in-grafana-mimir

Additionally, IIRC, I had a reasoning that one might want to keep specific certificates separately, and external env vars like `SSL_CERT_FILE` are not explicit enough and / or may not be available. This reasoning is debatable and a matter of product design, of course.

Hope that helps!

codebien: Thanks, @yorugac. If I'm understanding well we don't have a specific requirement for it; then it sounds reasonable to me to keep the k6 and the output config consistent and with as less as possible options for reducing the risk of breaking change in the future. I would be happy to bring back this if we will receive an explicit request for improving the UX on this front.

>if there is another way to connect to such an instance without using env var, it should be checked.

 There is and it's even better than env var because it is just implicit and it is the system trusted certs store. We are doing the same with the k6 core. https://github.com/grafana/k6/issues/218#issuecomment-301127630

yorugac: Yes, sounds good to me :+1:

PR #49, opened 2022-10-21
Removed the prometheus/prometheus go library

### LINKED ISSUE #41: Remove prometheus import
The [github.com/prometheus/prometheus](https://pkg.go.dev/github.com/prometheus/prometheus) Go package is not really required for this project.

The main reason is the HTTP client to deliver the series' samples and to get the protobuf definition.
We can create an internal HTTP client and import the protobuf types using the experimental [buf.build/prometheus/prometheus](https://buf.build/prometheus/prometheus/assets/main).

There is another minor usage related to timestamps' generation but it can replaced with the Go [time.UnixMilli](https://pkg.go.dev/time#Time.UnixMilli) method.

codebien:
It has been replaced with a lighter version based on [buf.build](https://buf.build/prometheus/prometheus/assets).

Closes #41


pkg/remotewrite/config.go
```diff
@@ -1,108 +1,85 @@
...
-
-	Url null.String `json:"url" envconfig:"K6_PROMETHEUS_REMOTE_URL"` // here, in the name of env variable, we assume that we won't need to distinguish between remote write URL vs remote read URL
-
+	// here, in the name of env variable, we assume that we won't need to distinguish between remote write URL vs remote read URL
+	URL     null.String       `json:"url" envconfig:"K6_PROMETHEUS_REMOTE_URL"`
 	Headers map[string]string `json:"headers" envconfig:"K6_PROMETHEUS_HEADERS"`
 
 	InsecureSkipTLSVerify null.Bool   `json:"insecureSkipTLSVerify" envconfig:"K6_PROMETHEUS_INSECURE_SKIP_TLS_VERIFY"`
-	CACert                null.String `json:"caCertFile" envconfig:"K6_CA_CERT_FILE"`
+	CACert                null.String `json:"caCertFile" envconfig:"K6_PROMETHEUS_CA_CERT_FILE"`
```
olegbespalov, line 31: Isn't that from another pull request? :thinking:

pkg/remote/client.go
```diff
@@ -0,0 +1,84 @@
...
+// the request is the proto marshalled and encoded.
+func (c *WriteClient) Store(ctx context.Context, series []*prompb.TimeSeries) error {
+	b, err := proto.Marshal(&prompb.WriteRequest{
+		Timeseries: series,
+	})
+	if err != nil {
+		return fmt.Errorf("encoding series as protobuf write request failed: %w", err)
+	}
+	b = snappy.Encode(nil, b) // TODO: this call can panic
+	req, err := http.NewRequest(http.MethodPost, c.url.String(), bytes.NewReader(b))
```
olegbespalov, line 63: Should we use `http.NewRequestWithContext` here?

pkg/remote/client.go
```diff
@@ -0,0 +1,84 @@
...
+// Store sends a batch of samples to the HTTP endpoint,
+// the request is the proto marshalled and encoded.
+func (c *WriteClient) Store(ctx context.Context, series []*prompb.TimeSeries) error {
+	b, err := proto.Marshal(&prompb.WriteRequest{
+		Timeseries: series,
+	})
+	if err != nil {
+		return fmt.Errorf("encoding series as protobuf write request failed: %w", err)
+	}
+	b = snappy.Encode(nil, b) // TODO: this call can panic
```
olegbespalov, line 62: :thinking: let's maybe put more details here, otherwise not super-useful TODO :smile:

pkg/remote/client.go
```diff
@@ -0,0 +1,84 @@
...
+	req, err := http.NewRequest(http.MethodPost, c.url.String(), bytes.NewReader(b))
+	if err != nil {
+		return fmt.Errorf("create new HTTP request failed: %w", err)
+	}
+
+	req.Header.Add("Content-Encoding", "snappy")
+	req.Header.Set("Content-Type", "application/x-protobuf")
+	req.Header.Set("User-Agent", "k6-prometheus-rw-output")
+	req.Header.Set("X-Prometheus-Remote-Write-Version", "0.1.0")
```
olegbespalov, line 71: This could easily be missed to be upgraded with the release, so maybe we need some variable Version as we have in the other projects. On the other side if we merge the extension to the k6 there should be a k6 version.

So up to you if we should change something here :)

pkg/remotewrite/remotewrite.go
```diff
@@ -39,19 +38,18 @@ func New(params output.Params) (*Output, error) {
...
+	clientConfig, err := config.RemoteConfig()
 	if err != nil {
 		return nil, err
 	}
 
-	// name is used to differentiate clients in metrics
-	client, err := remote.NewWriteClient("xk6-prwo", remoteConfig)
+	wc, err := remote.NewWriteClient(config.URL.String, clientConfig)
 	if err != nil {
 		return nil, err
```
olegbespalov, line 48: Maybe wrap an error here, to make it more clear to the end user, something like: `failed to initialize Prometheus remote write` or something? WDYT?

pkg/remote/client_test.go
```diff
@@ -0,0 +1,103 @@
...
+	})
+}
+
+func TestClientStore(t *testing.T) {
+	t.Parallel()
+	h := func(rw http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, r.Header.Get("Content-Encoding"), "snappy")
+		assert.Equal(t, r.Header.Get("Content-Type"), "application/x-protobuf")
+		assert.Equal(t, r.Header.Get("User-Agent"), "k6-prometheus-rw-output")
+		assert.Equal(t, r.Header.Get("X-Prometheus-Remote-Write-Version"), "0.1.0")
```
olegbespalov, line 50: It's probably okay to test that this header isn't empty. Otherwise, a test should be adjusted every release.

pkg/remote/client.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+type BasicAuth struct {
+	User, Password string
+}
+
+type WriteClient struct {
+	hc  *http.Client
+	url *url.URL
+	cfg *HTTPConfig
```
olegbespalov, line 33: :thinking: Maybe I'm missing something, but where do we actually use it? I mean, let's say the endpoint requires basic auth or TLS config, or even we'd like to pass the headers.

pkg/remote/client.go
codebien: This is some from the old code, now you're mentioning I try to directly fix or open an issue for this.

olegbespalov: I just spotted that we even mention this in the `README` https://github.com/grafana/xk6-output-prometheus-remote/blob/main/README.md?plain=1#L31

olegbespalov: Checked the code. Maybe we validate before encoding and return an error instead panic :thinking: 

Something like:
```golang
if snappy.MaxEncodedLen(len(b)) < 0 {
    return errors.New("to large to encode")
}
```

codebien: Yeah, it was exactly what I was going to do

pkg/remote/client_test.go
codebien: They are defined by the Promethus specs, I added a link to it

pkg/remote/client.go
codebien: Good catch, I forgot to include this.I will do.

pkg/remotewrite/remotewrite.go
codebien: Added in the latest fixup

pkg/remote/client.go
codebien: Same as for the test, it is defined from the protocol

pkg/remote/client.go
codebien: Added in the latest fixup

pkg/remote/client.go
codebien: Done

pkg/remotewrite/config.go
codebien: I expect we will merge the config PR soon so we will get this rid when I will rebase, just keep this open as a reminder.

pkg/remote/client_test.go
olegbespalov: Yeah, my fault. For some reason, I thought that this is about an extension :facepalm:

pkg/remotewrite/config.go
olegbespalov: Yeah, I was confused that this PR has only 1 commit :sweat_smile:

pkg/remote/client.go
```diff
@@ -0,0 +1,93 @@
...
+// the request is the proto marshalled and encoded.
+func (c *WriteClient) Store(ctx context.Context, series []*prompb.TimeSeries) error {
+	b, err := proto.Marshal(&prompb.WriteRequest{
+		Timeseries: series,
+	})
+	if err != nil {
+		return fmt.Errorf("encoding series as protobuf write request failed: %w", err)
+	}
+	if snappy.MaxEncodedLen(len(b)) < 0 {
+		return errors.New("the protobuf message is too large to be handled by Snappy encoder")
```
olegbespalov, line 67: Maybe also include the actual length for the error message, WDYT? :thinking:

pkg/remote/client.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/gogo/protobuf/proto"
+	"github.com/golang/snappy"
+	prompb "go.buf.build/grpc/go/prometheus/prometheus"
+)
+
+type HTTPConfig struct {
+	Timeout   time.Duration
+	TLSConfig *tls.Config
+	BasicAuth *BasicAuth
+	Headers   http.Header
```
olegbespalov, line 22: Should we pass headers with the request? :thinking:

pkg/remote/client.go
```diff
@@ -0,0 +1,109 @@
...
+	defer resp.Body.Close()
+	io.Copy(io.Discard, resp.Body) //nolint:errcheck
+
+	if resp.StatusCode != http.StatusNoContent {
+		return fmt.Errorf("got status code: %s instead expected: 204 No Content", resp.Status)
+	}
+	return nil
+}
+
+func (c *WriteClient) requestBody(series []*prompb.TimeSeries) ([]byte, error) {
```
olegbespalov, line 97: Let's maybe detach it from the `c *WriteClient`; as long as it can be a simple function, it's better to keep it as it is.

pkg/remotewrite/config.go
```diff
@@ -1,108 +1,85 @@
...
+	defaultMetricPrefix = "k6_"
 )
 
 type Config struct {
-	Mapping null.String `json:"mapping" envconfig:"K6_PROMETHEUS_MAPPING"`
-
-	Url null.String `json:"url" envconfig:"K6_PROMETHEUS_REMOTE_URL"` // here, in the name of env variable, we assume that we won't need to distinguish between remote write URL vs remote read URL
-
+	// here, in the name of env variable, we assume that we won't need to distinguish between remote write URL vs remote read URL
+	URL     null.String       `json:"url" envconfig:"K6_PROMETHEUS_REMOTE_URL"`
```
mstoykov, line 27: This comment makes me think there is some read functionality?

pkg/remotewrite/config.go
```diff
@@ -1,108 +1,85 @@
...
 	Headers map[string]string `json:"headers" envconfig:"K6_PROMETHEUS_HEADERS"`
 
 	InsecureSkipTLSVerify null.Bool   `json:"insecureSkipTLSVerify" envconfig:"K6_PROMETHEUS_INSECURE_SKIP_TLS_VERIFY"`
-	CACert                null.String `json:"caCertFile" envconfig:"K6_CA_CERT_FILE"`
+	CACert                null.String `json:"caCertFile" envconfig:"K6_PROMETHEUS_CA_CERT_FILE"`
+	User                  null.String `json:"username" envconfig:"K6_PROMETHEUS_USERNAME"`
```
mstoykov, line 32: nit: Maybe also rename the field to `Username`?

pkg/remotewrite/config.go
```diff
@@ -1,108 +1,85 @@
...
 )
 
 const (
-	defaultPrometheusTimeout = time.Minute
-	defaultFlushPeriod       = 5 * time.Second
-	defaultMetricPrefix      = "k6_"
+	defaultURL          = "http://localhost:9090/api/v1/write"
+	defaultTimeout      = 5 * time.Second
+	defaultPushInterval = 5 * time.Second
+	defaultMetricPrefix = "k6_"
```
mstoykov, line 22: From this I am left thinking you can configure both the metricPrefix and the timeout, but you can't. Maybe open issues about it?

mstoykov (approved):
LGTM :clap: 

I haven't tested it locally to see if it works, or there are any particular problems, but am also not accustomed with prometheus enough to find them :shrug: 


I have left a bunch of comments that are about renames and/or opening issues, but non of them are blocking the merge IMO

pkg/remote/client.go
codebien: I'm not sure this is a good idea. The function is returning a `prompb.WriteRequest` why should we use it outside of the `WriteClient` context? For example, if we detach then we have to rename with something like `writeRequestEncodedBody`.

pkg/remotewrite/config.go
codebien: It is, the Remote write protocol defines the endpoints for reading and for writing. Now, you're asking I'm thinking if we should break change and just have the path added as a constant from the `WriteClient`.

So the user instead passes:
```
K6_PROMETHEUS_REMOTE_URL=http://localhost:9090/api/v1/write
```
will pass:
```
K6_PROMETHEUS_REMOTE_URL=http://localhost:9090/api/v1/write
```

Potentially, we could also do both so use the explicit path if set otherwise add it but I don't know if we should support this complexity without an explicit feature request from users.

@grafana/k6-devs WDYT?

pkg/remotewrite/config.go
codebien: @mstoykov this is part of https://github.com/grafana/xk6-output-prometheus-remote/pull/48, can you review it there, please?

pkg/remotewrite/config.go
codebien: The question now is, do we want them? Because if we think they are required then I can just fix them in https://github.com/grafana/xk6-output-prometheus-remote/pull/48.

Instead, if we think we should wait for users then I will open an issue and add a comment for it in the code. (I would prefer this option for keeping the extension as light as possible).

pkg/remote/client.go
olegbespalov: Because at least now, it requires nothing from the actual `*WriteClient`. Even if the function returns encoded `prompb.WriteRequest` that doesn't mean that it should be in responsibility of the `WriteClient`. Simply it shouldn't be responsible for everything related to the writing :shrug: 

>For example, if we detach then we have to rename with something like `writeRequestEncodedBody`.

I see your point, but I can't agree that it sounds to everybody that if a `requestBody` method belongs to the `WriteClient` automatically helps to understand what it is without knowing the exact place where we use it or looking into its body. Another word I can understand is that it actually returns an encoded `prompb.WriteRequest` I either can get by a reading of the method or only together with the invocation place:
```
func (c *WriteClient) Store(ctx context.Context, series []*prompb.TimeSeries) error {
	b, err := c.requestBody(series)
	if err != nil {
		return err
	}
        ...
```

`writeRequestEncodedBody` sounds good to me, to be honest, but I'm fine even with the existing method name as long as it's the single request that the client is doing. Maybe just to reserve some place for the future and follow go conventions name it in a constructor like style `newWriteRequestBody()`.

pkg/remotewrite/config.go
olegbespalov: @codebien, But how likely is it possible that in this output extension, we will ever use the read endpoint? :thinking: I agree with this, and it seems that this comment can probably lead to some confusion, and its value is questionable.

>WDYT?

IMO, the first option (where we define the endpoint `http://localhost:9090/api/v1/write`) still looks a bit more flexible and preferable as long as we not planning to use anything apart from that endpoint.

pkg/remotewrite/config.go
mstoykov: I think opening issues for now is sufficient - we can always add them later

olegbespalov (approved):
Generally LGTM :+1: Great work! :muscle:

go.mod
```diff
@@ -1,58 +1,36 @@
 module github.com/grafana/xk6-output-prometheus-remote
 
-go 1.17
+go 1.18
 
 require (
-	github.com/golang/protobuf v1.5.2
+	github.com/gogo/protobuf v1.3.2
```
olegbespalov, line 6: The project is deprecated. There is even a youtube video around it https://www.youtube.com/watch?v=HTIltI0NuNg

Not sure if we need it to do in this PR, but at some point, we should migrate out of it :sweat_smile:

olegbespalov: If it's not a big deal, maybe we can revert this change in this PR not to produce new tasks.

pkg/remotewrite/config.go
codebien: #52 #51

pkg/remotewrite/config.go
codebien: I rechecked the specs and the path is not fixed, I was convinced the protocol defined the path but this is obviously not true. The path depends on the implementation and it is configurable so we have to get it explicitly.

codebien: I added documentation for these fields so we can now remove the original comment.

pkg/remote/client.go
codebien: Changed and added a test for it

pkg/remote/client.go
```diff
@@ -0,0 +1,113 @@
...
+
+type HTTPConfig struct {
+	Timeout   time.Duration
+	TLSConfig *tls.Config
+	BasicAuth *BasicAuth
+	Headers   http.Header
+}
+
+type BasicAuth struct {
+	User, Password string
```
olegbespalov, line 26: non-blocking-nit: if we rename config ones to `User` it is also good to keep that synced

olegbespalov (approved):
LGTM, left one non-blocking comment! :+1:

mstoykov (approved):

pkg/remote/client.go
codebien: Added, good catch, thanks

codebien: I renamed to `Username`

PR #55, opened 2022-11-08
Migrate to k6 v0.41

codebien:
The minimal refactor for migrating to k6 v0.41 and reusing its generated time series. More refactoring would be useful but I plan to do it in a second step supporting it with benchmarks.


mstoykov (approved):

pkg/remotewrite/prometheus.go
```diff
@@ -4,21 +4,26 @@ import (
...
 )
 
-func MapTagSet(t *metrics.SampleTags) []*prompb.Label {
-	tags := t.CloneTags()
-
-	labels := make([]*prompb.Label, 0, len(tags))
-	for k, v := range tags {
-		labels = append(labels, &prompb.Label{Name: k, Value: v})
+func MapTagSet(t *metrics.TagSet) []*prompb.Label {
+	n := (*atlas.Node)(t)
```
olegbespalov, line 13: Shouldn't we use `t.Map()`? :thinking:

codebien: No, t.Map generates a map[string]string. Instead, going up to the tree we are reusing as much memory as possible.

codebien: Going to merge, let me know if you think we should change.

olegbespalov: I see your point, but what I don't feel comfortable with is that we rely on internals (we're exposing the `*atlas.Node`) :cry: 

How about extending the `metrics.TagSet` with something like:
```golang
func (ts *TagSet) Len() int {
	return (*atlas.Node)(ts).Len()
}

// Iterate memory efficient way of iterating all tags
func (ts *TagSet) Iterate(fn func(key, value string)) {
	if ts.Len() < 1 {
		return
	}

	n := (*atlas.Node)(ts)

	for !n.IsRoot() {
		prev, key, value := n.Data()
		fn(key, value)
		n = prev
	}
}
```

And then, the body of this method will look like
```golang
labels := make([]*prompb.Label, 0, t.Len())
t.Iterate(func (key, value string) {
   labels = append(labels, &prompb.Label{Name: key, Value: value})
}) 
return labels
```

@mstoykov @codebien  WDYT? Could be useful for many cases?

mstoykov: This whole extension is destined to become internal - so refactoring this seems not very useful IMO. 

Can this be used by other outputs - likely yes, so go ahead and open a k6 issue, but lets not block the work on the extension on this.

olegbespalov: No, I do not intend to block the PR; in any case, it's merged.

 >This whole extension is destined to become internal - so refactoring this seems not very useful IMO.

Here I beg to differ. Even though this code becomes a part of the k6, which makes it an internal module, that relation between the packages still won't be internal, so having (and using) a clean code is better. 

Moreover, the code itself remains an example; ideally, the example should be good enough for the others.

na--: IIRC, I actually intended to add something like the `Iterate()` method @olegbespalov described in `TagSet`, just never got around to it. So :+1: for having something like that in k6 and not needing to cast things to `*atlas.Node`

codebien: I find it valuable, I already suggested it in the past but we haven't hit the requirement for it before. It could require some profiling for being sure we are not allocating unexpected memory. @olegbespalov Do you want open a PR on k6 or would you prefer that I do it?

olegbespalov: As @mstoykov suggested, I've opened the issue https://github.com/grafana/k6/issues/2764 for now to keep track.

If you want to open a PR go ahead :+1:

olegbespalov: Feel free to put your thoughts (like the @codebien 's suggestion about memory profiling which I found valuable) there

PR #57, opened 2022-11-16
Experimental Native Histogram

codebien:
It adds an experimental mapping for Trend. It uses the new Prometheus' experimental feature Native Histogram.
A new `TrendAsNativeHistogram` option has been added for enabling it. If it is not explicitly enabled the Output continues to use the current AsGauges mapping.

It also contains other changes not dedicated to this feature so they will be separated in the next few days with specific PRs:

- Sort labels
- Migration to k6.metrics.TrendSink (from the master branch)

#### Manual testing

The Docker compose has been updated so it can be used for testing. A new xk6 build based on this branch is required.

```sh
$ docker compose up -d prometheus grafana
```

```sh
K6_PROMETHEUS_TREND_AS_NATIVE_HISTOGRAM=true ./k6 run -o output-prometheus-remote ./samples/test.js
```
An example PromQL query for checking the generated data:
```sh
histogram_quantile(0.9, rate(k6_http_reqs_duration[1m]))
```


pkg/remotewrite/trend.go
```diff
@@ -1,84 +1,66 @@
...
 
+	"github.com/prometheus/client_golang/prometheus"
 	"go.k6.io/k6/metrics"
 )
 
-// The following functions are an attempt to add ad-hoc optimization to TrendSink,
-// and are a partial copy-paste from k6/metrics.
-// TODO: re-write & refactor this once metrics refactoring progresses in k6.
-
 type trendSink struct {
```
olegbespalov, line 10: What do you think if we split implementations? 

Like current terndSink can be an old version and introduce another `trendNativeHistogram.` 

That way, we can avoid this all `if t.H == nil { ... return }`, which makes code more straightforward.

pkg/remotewrite/remotewrite.go
```diff
@@ -41,16 +42,24 @@ func New(params output.Params) (*Output, error) {
 		return nil, err
 	}
 
+	var tm trendMapper
```
olegbespalov, line 45: If I got the code properly, that also can be a part of the separation (I mean that particular can be part of the sink structures) that I mentioned an in another comment and somewhere bellow you can call somewhere bellow: 

instead of
```golang
newts = swm.TrendMapper.Map(swm.TimeSeries, swm.Latest, trend)
```
Like that:
```
newts = trend.Map(swm.TimeSeries, swm.Latest)
```

pkg/remotewrite/prometheus.go
```diff
@@ -141,3 +149,56 @@ func MapTrend(series metrics.TimeSeries, t time.Time, sink *trendSink) []*prompb
...
+func MapTrendAsNativeHistogram(series metrics.TimeSeries, t time.Time, sink *trendSink) []*prompb.TimeSeries {
+	labels := MapTagSet(series.Tags)
+	timestamp := t.UnixMilli()
+
+	return []*prompb.TimeSeries{
+		{
+			// TODO: should we map the builtin metrics
+			// with the Prometheus' convention?
+			//
+			// e.g. k6_http_reqs_duration_seconds
```
olegbespalov, line 163: I wish I'd say that we should follow the Prometheus convention. Still, on the other side, I think we have a lot of the documentation that remains referring to it as the `k6_http_reqs_duration`, so if we adjust it, we should somehow explain it in other places, so not sure what we should do, but maybe raise it in a wider-team discussion.

codebien: >so if we adjust it, we should somehow explain it in other places

Just for clarification, this would be only for the Prometheus Output, so it would be visible only from a Prometheus server (or any other compatible storage). In k6 it would continue to remain the same current name.

So, do you have some examples of other places where you would like to mention it?

olegbespalov: That's what I got, but still, if we change the name, it should be communicated. I know that we also attach prefixes, but generally, the name change looks too big for me :shrug: 

Maybe I'm over-thinking there :sweat_smile:

codebien: Ofc, we can communicate them in the release notes of the new version. I don't see other places where we could announce them. Also, this is an extension so there isn't a stability guarantee.

olegbespalov: Okay, then let's consider that I don't have a strong opinion here apart from the explicitly communicating and documenting what ever will be result :+1:

pkg/remotewrite/remotewrite.go
codebien: Thinking better on this, I think this must be refactored. The current design generates a mapper for each series. We should try reversing the pattern, we should allocate one mapper and pass seriesWithMeasure to it. I will try to change it.

olegbespalov: > The current design generates a mapper for each series. We should try reversing the pattern, we should allocate one mapper and pass seriesWithMeasure to it

I'm not sure if I got this :thinking: Is the mapper itself a heavy struct?

Just in case, I see no issues with having different mappers. It's absolutely okay when we have complex data structures.

What I'm saying is that in the case of that PR `Map` function belongs to the sink's structs. Basically, in that case, sync will just satisfy the `Mapper` interface, and we don't need somewhere also keep the correct reference as we do now `output.trendMapper`

codebien:
@olegbespalov I have rebased and addressed the request changes

pkg/remotewrite/trend.go
```diff
@@ -107,3 +110,108 @@ func (tg *trendAsGauges) CacheNameIndex() {
...
+	}
+}
+
+// TODO: create a smaller Sink interface for this Output.
+// Sink with only Add and MapPrompb methods should be enough.
+// One method interfaces could be even better, to be checked.
+
+// P implements metrics.Sink.
+func (*nativeHistogramSink) P(pct float64) float64 {
+	panic("P shouldn't be invoked when Native Histogram mapping is used")
```
olegbespalov, line 156:
```suggestion
	panic("Native Histogram has no support of percentile (P) ")
```

pkg/remotewrite/trend.go
```diff
@@ -107,3 +110,108 @@ func (tg *trendAsGauges) CacheNameIndex() {
...
+}
+
+// Format implements metrics.Sink.
+func (*nativeHistogramSink) Format(td time.Duration) map[string]float64 {
+	panic("Format shouldn't be invoked when Native Histogram mapping is used")
```
olegbespalov, line 161:
```suggestion
	panic("Native Histogram has no support of formatting (Format)")
```

pkg/remotewrite/trend.go
```diff
@@ -107,3 +110,108 @@ func (tg *trendAsGauges) CacheNameIndex() {
...
+}
+
+// IsEmpty implements metrics.Sink.
+func (*nativeHistogramSink) IsEmpty() bool {
+	panic("IsEmpty shouldn't be invoked")
```
olegbespalov, line 166:
```suggestion
	panic("Native Histogram has no support of emptiness check (IsEmpty)")
```

pkg/remotewrite/config.go
```diff
@@ -224,6 +236,15 @@ func GetConsolidatedConfig(jsonRawConf json.RawMessage, env map[string]string, a
 		result.Headers[k] = v
 	}
 
+	if b, err := getEnvBool(env, "K6_PROMETHEUS_TREND_AS_NATIVE_HISTOGRAM"); err != nil {
+		return result, err
+	} else {
+		if b.Valid {
+			// apply only if valid, to keep default option otherwise
```
olegbespalov, line 243:
```suggestion
```
It's pretty clear from the code :wink:

pkg/remotewrite/trend_test.go
```diff
@@ -83,3 +85,61 @@ func TestTrendAsGaugesFindIxName(t *testing.T) {
...
+		Tags: r.RootTagSet().With("tagk1", "tagv1"),
+	}
+
+	sample := metrics.Sample{
+		TimeSeries: series,
+		Value:      1.52,
+		Time:       time.Unix(1, 0),
+	}
+	st := newNativeHistogramSink(sample.Metric)
+	st.Add(sample)
```
olegbespalov, line 133: Maybe, just in case, add more than one sample, WDYT?

olegbespalov (commented):
Generally LGTM :muscle: 

I left a few minor suggestions about the panics. The idea is to make them more transparent for the facing user.

Last but not least. Do you think we should update the README with the info about `K6_PROMETHEUS_TREND_AS_NATIVE_HISTOGRAM` :thinking:

pkg/remotewrite/trend.go
```diff
@@ -107,3 +110,108 @@ func (tg *trendAsGauges) CacheNameIndex() {
...
+			},
+		},
+	}
+}
+
+func histogramToHistogramProto(timestamp int64, h prometheus.Histogram) *prompb.Histogram {
+	// TODO: research more if a better way is possible.
+	metric := &dto.Metric{}
+	if err := h.Write(metric); err != nil {
+		panic(fmt.Errorf("conversion from internal Metric model to Proto failed: %w", err))
```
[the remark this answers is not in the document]
codebien: In that place is already a Native Histogram. The conversion is between a Native Histogram and a Protobuf Native Histogram.

```suggestion
		panic(fmt.Errorf("failed to convert Native Histogram to the related Protobuf: %w", err))
```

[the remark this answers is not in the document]
olegbespalov: Yeah, right :+1:

olegbespalov (approved):
:+1:

PR #58, opened 2022-11-18
TrendSink from k6 and sorted labels

codebien:
Refactoring that contains the following updates:
* Now, it uses directly the TrendSink defined from the metrics pkg of k6. The recent refactor on k6 has enabled this step. It has required bumping the k6 version to the master branch.
*  The generated labels are now sorted in a lexicographic order as required by the Remote write specification.

### DIFF
--- a/pkg/remotewrite/trend.go
+++ b/pkg/remotewrite/trend.go
@@ -33,7 +33,7 @@ func (sink *extendedTrendSink) MapPrompb(series metrics.TimeSeries, t time.Time)
 		labels:    MapSeries(series),
 		timestamp: t.UnixMilli(),
 	}
-	tg.FindNameIndex()
+	tg.CacheNameIndex()
 
 	tg.Append("count", float64(sink.Count))
 	tg.Append("sum", sink.Sum)
@@ -85,11 +85,11 @@ func (tg *trendAsGauges) Append(suffix string, v float64) {
 	tg.series = append(tg.series, ts)
 }
 
-// FindNameIndex finds the __name__ label's index
+// CacheNameIndex finds the __name__ label's index
 // if it is different from the most common expected case
 // then it caches the value.
 // The labels slice is expected to be sorted.
-func (tg *trendAsGauges) FindNameIndex() {
+func (tg *trendAsGauges) CacheNameIndex() {
 	if tg.labels[0].Name == namelbl {
 		// ixname is expected to be the first in most of the cases
 		// the default value is already 0
--- a/pkg/remotewrite/trend_test.go
+++ b/pkg/remotewrite/trend_test.go
@@ -79,7 +79,7 @@ func TestTrendAsGaugesFindIxName(t *testing.T) {
 			lbls = append(lbls, &prompb.Label{Name: l})
 		}
 		tg := trendAsGauges{labels: lbls}
-		tg.FindNameIndex()
+		tg.CacheNameIndex()
 		assert.Equal(t, tc.expIndex, tg.ixname)
 	}
 }


pkg/remotewrite/prometheus.go
```diff
@@ -23,121 +29,109 @@ func MapTagSet(t *metrics.TagSet) []*prompb.Label {
...
+		Value:     v,
+	}
+	tg.series = append(tg.series, ts)
+}
+
+// FindNameIndex finds the __name__ label's index
+// if it is different from the most common expected case
+// then it caches the value.
+// The labels slice is expected to be sorted.
+func (tg *trendAsGauges) FindNameIndex() {
```
olegbespalov, line 120: I'd recommend giving another name. The `FindNameIndex` confuses the main purpose of the method as I see is to cache if possible, so my alternatives are something like: `BuildNameIndexCache`, `TryCacheNameIndex`, or something which explicitly tells the purpose of the method.

pkg/remotewrite/prometheus.go
```diff
@@ -23,121 +29,109 @@ func MapTagSet(t *metrics.TagSet) []*prompb.Label {
...
+		return lbls[i].Name < lbls[j].Name
+	})
+	return lbls
 }
 
-func MapTrend(series metrics.TimeSeries, t time.Time, sink *trendSink) []*prompb.TimeSeries {
+// MapTrendAsGauges converts a k6 time series and its relative
+// Sink into the equivalent TimeSeries model as defined from
+// the Remote write specification.
+func MapTrendAsGauges(series metrics.TimeSeries, t time.Time, sink *metrics.TrendSink) []*prompb.TimeSeries {
```
olegbespalov, line 52: What are our plans for the neighbor PR #57?

I'm asking because if we want to continue to work on it, it make sense to implement https://github.com/grafana/xk6-output-prometheus-remote/pull/57/files#r1025272534

codebien: I will try to do it here, I added a comment directly in #57. I think there is the risk to expand too much the scope in this PR. If I will add too many changes then I will submit a dedicated PR on top of this. @olegbespalov WDYT?

olegbespalov: That's why I'm confused a bit. I guess these two PRs are changing the same parts of the logic/structs, which will lead to conflict. Maybe we can combine them into one with different commits?

codebien: Sorry, I thought it was obvious but it was a bad assumption. I'm splitting from #57 all the things not strictly related to Histogram. So we can focus there mostly on reviewing that part.

olegbespalov (approved):
:+1:

pkg/remotewrite/prometheus.go
olegbespalov: :+1:

pkg/remotewrite/trend.go
```diff
@@ -1,84 +1,109 @@
...
-	default:
-		// If percentile falls on a value in Values slice, we return that value.
-		// If percentile does not fall on a value in Values slice, we calculate (linear interpolation)
-		// the value that would fall at percentile, given the values above and below that percentile.
-		i := pct * (float64(t.Count) - 1.0)
-		j := t.Values[int(math.Floor(i))]
-		k := t.Values[int(math.Ceil(i))]
-		f := i - math.Floor(i)
-		return j + (k-j)*f
+// FindNameIndex finds the __name__ label's index
```
olegbespalov, line 88:
```suggestion
// CacheNameIndex finds the __name__ label's index
```

codebien: Apparently, we have something to fix in the linter...

PR #62, opened 2022-11-24
Name suffix convention for Native Histograms

codebien:


pkg/remotewrite/prometheus.go
```diff
@@ -34,11 +34,14 @@ func MapTagSet(t *metrics.TagSet) []*prompb.Label {
...
 // The labels are lexicographic sorted as required
 // from the Remote write's specification.
-func MapSeries(series metrics.TimeSeries) []*prompb.Label {
+func MapSeries(series metrics.TimeSeries, suffix string) []*prompb.Label {
 	lbls := append(MapTagSet(series.Tags), &prompb.Label{
 		Name:  namelbl,
 		Value: fmt.Sprintf("%s%s", defaultMetricPrefix, series.Metric.Name),
 	})
+	if suffix != "" {
+		lbls[len(lbls)-1].Value += "_" + suffix
```
olegbespalov, line 43: :thinking: Out of curiosity, do we need to set it on an array with `lbls[len(lbls)-1]` ?

Could it be more straightforward, something like:
```
v := defaultMetricPrefix + series.Metric.Name
if suffix != "" {
    v+= suffix
}

...
// use v

File: pkg/remotewrite/trend.go [hunk 2]
```diff
@@ -215,3 +214,14 @@ func toBucketSpanProto(s []*dto.BucketSpan) []*prompb.BucketSpan {
 	}
 	return spans
 }
+
+func baseUnit(vt metrics.ValueType) string {
```
olegbespalov, line 218: What if we call it `unitSuffix` and return the actual suffix (`_seconds` and `_bytes`),  then later we don't need any condition like `if suffix != ""` but, just `v := defaultMetricPrefix + series.Metric.Name + suffix`

codebien: I'm not getting how you would manage the case when the ValueType is Undefined (so not Time nor Data).

olegbespalov: just return an empty sting (a.k.a. null suffix :sweat_smile: )

codebien: >if suffix != "" but, just v := defaultMetricPrefix + series.Metric.Name + suffix

As shared in private, this logic is applied from `MapSeries` which is a general function where the baseUnit is one of the possible options for the suffix but not the unique.

olegbespalov (approved):
:+1:

PR #64, opened 2022-11-25
Configurable Trend stats

codebien:


pkg/remotewrite/trend_test.go
```diff
@@ -32,22 +32,26 @@ func TestExtendedTrendSinkMapPrompb(t *testing.T) {
...
 		buildTimeSeries("k6_test_count", 1.0, now),
-		buildTimeSeries("k6_test_sum", 1.0, now),
-		buildTimeSeries("k6_test_min", 1.0, now),
 		buildTimeSeries("k6_test_max", 1.0, now),
-		buildTimeSeries("k6_test_avg", 1.0, now),
 		buildTimeSeries("k6_test_med", 1.0, now),
+		buildTimeSeries("k6_test_min", 1.0, now),
 		buildTimeSeries("k6_test_p90", 1.0, now),
 		buildTimeSeries("k6_test_p95", 1.0, now),
+		// buildTimeSeries("k6_test_sum", 1.0, now),
```
codebien, line 42: https://github.com/grafana/k6/pull/2785

pkg/remotewrite/config.go
```diff
@@ -259,6 +277,16 @@ func parseArg(text string) (Config, error) {
...
+
+		// TODO: add the support for trendStats
+		// strvals doesn't support the same format used by --summary-trend-stats
+		// using the comma as the separator, because it is already used for
+		// dividing the keys.
+		//
+		//if v, ok := params["trendStats"].(string); ok && len(v) > 0 {
+		//c.TrendStats = strings.Split(v, ",")
+		//}
+
```
codebien, line 289: @na-- Ideas? Should we use another char (e.g. `|`)?

```
./k6 -o prometheus-remote-write url=http://promurl/api/v1/write,trendStats=p(99)|avg
```

WDYT?

na--: hmm no, sorry, but this might be a good time to re-evaluate if we want to use something like strvals at all :thinking: 

IIRC this was one of the reason that the influxdb output decodes text parameters as a URL, though that also has its issues :disappointed:

na--: we can postpone this decision by not supporting any text-based configuration at the beginning :thinking: just environment variables will probably be sufficient, for now?

codebien: https://github.com/grafana/xk6-output-prometheus-remote/pull/70

pkg/remotewrite/remotewrite.go
```diff
@@ -76,6 +85,50 @@ func (o *Output) Stop() error {
...
+	o.trendStatsResolver = make(TrendStatsResolver, len(resolvers))
+	for stat, fn := range resolvers {
+		statKey := stat
+
+		// the config passes percentiles with p(x) form, for example p(95),
+		// but the mapping generates series name in the form p95.
+		//
+		// TODO: maybe decoupling mapping from the stat resolver keys?
+		if strings.HasPrefix(statKey, "p(") {
+			statKey = stat[2 : len(statKey)-1]             // trim the parenthesis
```
oleiade, line 123: I believe you could use[ `strings.TrimPrefix`](https://pkg.go.dev/strings#TrimPrefix) here instead 👍🏻

pkg/remotewrite/remotewrite_test.go
```diff
@@ -220,3 +221,55 @@ func TestNewSeriesWithNativeHistogramMeasure(t *testing.T) {
...
+	for _, tt := range tests {
+		o := Output{}
+		err := o.setTrendStatsResolver(tt.stats)
+		require.NoError(t, err)
+		require.NotNil(t, o.trendStatsResolver)
+
+		assert.Len(t, o.trendStatsResolver, len(tt.expResolverKeys))
+		assert.ElementsMatch(t, tt.expResolverKeys, func() []string {
+			var keys []string
+			for statKey := range o.trendStatsResolver {
```
oleiade, line 269: I think you could use `copy` here instead. That would make the code even shorter?

oleiade (approved):
Some tiny comments, but looks really good overall, great work 👍🏻

codebien:
@oleiade thanks, your request changes are internal changes, so I'm going to merge then address them in a dedicated PR.

oleiade:
@codebien makes sense 🙇🏻

pkg/remotewrite/config.go
```diff
@@ -21,6 +21,8 @@ const (
 	defaultMetricPrefix = "k6_"
 )
 
+var defaultTrendStats = []string{"p(99)"}
+
```
codebien, line 25: @na-- @oleiade it would be good to get an opinion on this. I took this decision autonomously, but you could have a different opinion.

na--: For me it makes some sense that the default value should match whatever the value of `--summary-trend-stats` in k6 is (default or user-supplied). But I can also see why that might be a bit too expensive to maintain by default... So, I don't have a strong opinion either way, sorry.

PR #70, opened 2022-11-29
Drop options from cli argument

codebien:
k6 doesn't have a defined way of providing Output options. Instead of adding one more way, we should define one standard for all of them and only after adopting it here.


javaducky:
Agreed...we need to come up with a good option or simply utilize env vars. I dislike how we have URLs tagging along on the output type for TimescaleDB (e.g. `-o timescaledb=postgresql://k6:k6@timescaledb:5432/k6`) and InfluxDB (e.g. `-o xk6-influxdb=http://localhost:8086`) as well...reminds me too much of JDBC connection strings. 🤷

mstoykov (approved):

PR #72, opened 2022-11-30
Trend stats in base unit

codebien:
The previous stats emits the value as milliseconds, instead, this PR uniforms to the Prometheus convention to emit metrics' value in a base time unit.


imiric (approved):
:+1:

Issue #37, opened 2022-08-31
Incorrect data saved to prometheus


Kostoni:
Hi

I have a problem with prometheus. When I run script with saving metrics to prometheus I see aproximately 5 min of fake data at the end of test run metrics. 

It repeats on your sample from readme file. If you go to prometheus after test.js script execution you'll see that you have 5 minutes of data after the data for 10 seconds your test script saved.

![2022-08-31_16-56-33](https://user-images.githubusercontent.com/5243183/187680543-8636db03-b38d-49ed-a5a7-a68e291526bc.png)

You can see this fake data also using grafana dashboard from your example.

![2022-08-31_17-02-33](https://user-images.githubusercontent.com/5243183/187680562-7969edc1-04a6-4732-8b56-f745d773190d.png)

Repeated in 0.38, 0.39 version of k6

Thank you for your job!

Kostoni:
Hi again, we've found that reason was in flag query.lookback-delta with default value 5m. I think need to add this prometheus behavior to readme file and fix docker-compose file to create prometheus image with correct behavior. 

This fix should be enough to correct work.

`command: --web.enable-remote-write-receiver --query.lookback-delta=1s --config.file=/etc/prometheus/prometheus.yml`

Thanks!

codebien:
This should be automatically fixed when https://github.com/grafana/xk6-output-prometheus-remote/issues/68 is added.

PR #73, opened 2022-11-30
Stale markers

### LINKED ISSUE #68: Stale markers
Add a Stale marker for each seen time series at the end of the test. This is required from the specification: https://docs.google.com/document/d/1LPhVRSFkGNSuU1fBd81ulhsCPR4hkSZyyBj1SZ8fWOM/edit#heading=h.hfqkr2527w2g

codebien:
Added a Stale marker at the end of the test for all the seen time series.  

Closes #68 #37

### DIFF
--- a/pkg/remotewrite/remotewrite.go
+++ b/pkg/remotewrite/remotewrite.go
@@ -3,6 +3,7 @@ package remotewrite
 import (
 	"context"
 	"fmt"
+	"math"
 	"strings"
 	"time"
 
@@ -15,7 +16,15 @@ import (
 	prompb "go.buf.build/grpc/go/prometheus/prometheus"
 )
 
-var _ output.Output = new(Output)
+var (
+	_ output.Output = new(Output)
+
+	// staleNaN is the Prometheus special value for marking
+	// a time series as stale.
+	//
+	// https://pkg.go.dev/github.com/prometheus/prometheus/pkg/value#pkg-constants
+	staleNaN = math.Float64frombits(0x7ff0000000000002)
+)
 
 type Output struct {
 	output.SampleBuffer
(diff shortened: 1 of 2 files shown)


pkg/remotewrite/remotewrite.go
```diff
@@ -15,7 +16,15 @@ import (
...
 
-var _ output.Output = new(Output)
+var (
+	_ output.Output = new(Output)
+
+	// staleNaN is the Prometheus special value for marking
+	// a time series as stale.
+	//
+	// https://pkg.go.dev/github.com/prometheus/prometheus/pkg/value#pkg-constants
+	staleNaN = math.Float64frombits(0x7ff0000000000002)
```
oleiade, line 26: Would it make sense to make this particular value a constant?

oleiade (approved):
LGTM 👏🏻 🎅🏻 🎄

pkg/remotewrite/remotewrite.go
codebien: I'm looking into this but I'm not sure that the alternative is really better because the const would be a uint64 and it would be accepted if assigned to a float64. So if we forget to convert with `math.Float64frombits(` then we introduce a bug because the compiler doesn't complain. We should only rely on unit tests. @oleiade WDYT?

pkg/remotewrite/remotewrite.go
```diff
@@ -80,11 +89,51 @@ func (o *Output) Start() error {
...
+
+	staleMarkers := o.staleMarkers(time.Now())
+	o.logger.WithField("staleMarkers", len(staleMarkers)).Debug("Marking time series as stale, if any")
+	if len(staleMarkers) < 1 {
+		return nil
+	}
+
+	// TODO: warn if it the request is taking too long (5 seconds?, the same timeout set for flushing)
+
+	err := o.client.Store(context.Background(), staleMarkers)
```
imiric, line 103: Hhmm shouldn't this use some extension-wide context? Preferably one passed from k6? I now see that `output.Params` doesn't have a context. Maybe it should?

pkg/remotewrite/remotewrite.go
```diff
@@ -80,11 +89,51 @@ func (o *Output) Start() error {
...
+	}
 	return nil
 }
 
+// staleMarkers maps all the seen time series with a stale marker.
+func (o *Output) staleMarkers(t time.Time) []*prompb.TimeSeries {
+	timestamp := t.UnixMilli()
+	staleMarkers := make([]*prompb.TimeSeries, 0, len(o.tsdb))
+
+	for _, swm := range o.tsdb {
```
imiric, line 115: Could this race with any writes to `o.tsdb`?

imiric (commented):
LGTM, but I'm not very familiar with the codebase yet, so pardon the possibly dumb questions :sweat_smile:

pkg/remotewrite/remotewrite.go
codebien: >Can't we just use the StaleNaN constant from the prometheus package directly?

@imiric It would require including the Prometheus project as a direct dependency, but it has a lot of dependencies.

pkg/remotewrite/remotewrite.go
codebien: `tsdb` is used by `flushMetrics` and `Stop`, `Stop` calls `o.periodicFlusher.Stop()` that should guarantee no more invocations for `flushMetrics`, so I don't think it can be racy.

The `periodicFlusher` core code:

https://github.com/grafana/k6/blob/4e4e5b1ff730ce4fcce8f8296b8fbb767ec429d9/output/helpers.go#L63-L86

pkg/remotewrite/remotewrite.go
codebien: >For this version, I think the unique way we have is to implement the TODO directly now logging in case of a long Stop execution. WDYT?

Thinking better on this, the current client is already initialized with the Timeout value from the configuration, so the request timed out when the timeout interval is passed. It should be enough for warning in the case of a long-time request and to alarm for a required investigation into eventual bad performance between k6 and the Remote write endpoint.

pkg/remotewrite/remotewrite.go
```diff
@@ -80,11 +89,51 @@ func (o *Output) Start() error {
 
 func (o *Output) Stop() error {
 	o.logger.Debug("Stopping the output")
+	defer o.logger.Debug("Output stopped")
 	o.periodicFlusher.Stop()
-	o.logger.Debug("Output stopped")
+
+	staleMarkers := o.staleMarkers(time.Now())
+	o.logger.WithField("staleMarkers", len(staleMarkers)).Debug("Marking time series as stale, if any")
```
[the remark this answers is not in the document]
codebien: Done with the fixup

pkg/remotewrite/remotewrite.go
imiric: Thanks for confirming :+1:

imiric (approved):
Nicely done :+1:

pkg/remotewrite/remotewrite.go
codebien: >Intuitively, it makes sense to me for the extension context to be based on one received by k6, so that the extension code can cancel any operations in progress if the k6 context is cancelled (due to a signal, test abort, etc.).

Just to be clear, I agree here, and this is something we want to add in the future, this is why the context is passed to the `wait` function in the [Output refactor proposal](https://github.com/grafana/k6/issues/2430#issuecomment-1073803076) `(wait func(context.Context) error, err error)`.

Unfortunately, I think we can't do it now, because it would require some changes to k6 when the code is almost frozen.

oleiade (approved):

PR #79, opened 2022-12-06
Rename from prometheus-remote to prometheus-rw

codebien:
It renames the registered name from `"output-prometheus-remote` to `xk6-prometheus-rw` and it makes the env vars consistent with the new name renaming the prefix to `K6_PROMETHEUS_RW_*`

### DIFF
--- a/pkg/remotewrite/config.go
+++ b/pkg/remotewrite/config.go
@@ -25,32 +25,32 @@ var defaultTrendStats = []string{"p(99)"}
 
 type Config struct {
 	// URL contains the absolute URL for the Write endpoint where to flush the time series.
-	URL null.String `json:"url" envconfig:"K6_PROMETHEUS_REMOTE_URL"`
+	URL null.String `json:"url" envconfig:"K6_PROMETHEUS_RW_REMOTE_URL"`
 
 	// Headers contains additional headers that should be included in the HTTP requests.
-	Headers map[string]string `json:"headers" envconfig:"K6_PROMETHEUS_HEADERS"`
+	Headers map[string]string `json:"headers" envconfig:"K6_PROMETHEUS_RW_HEADERS"`
 
 	// InsecureSkipTLSVerify skips TLS client side checks.
-	InsecureSkipTLSVerify null.Bool `json:"insecureSkipTLSVerify" envconfig:"K6_PROMETHEUS_INSECURE_SKIP_TLS_VERIFY"`
+	InsecureSkipTLSVerify null.Bool `json:"insecureSkipTLSVerify" envconfig:"K6_PROMETHEUS_RW_INSECURE_SKIP_TLS_VERIFY"`
 
 	// Username is the User for Basic Auth.
-	Username null.String `json:"username" envconfig:"K6_PROMETHEUS_USERNAME"`
+	Username null.String `json:"username" envconfig:"K6_PROMETHEUS_RW_USERNAME"`
 
 	// Password is the Password for the Basic Auth.
-	Password null.String `json:"password" envconfig:"K6_PROMETHEUS_PASSWORD"`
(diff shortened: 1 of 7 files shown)


pkg/remotewrite/config.go
```diff
@@ -25,32 +25,32 @@ var defaultTrendStats = []string{"p(99)"}
 
 type Config struct {
 	// URL contains the absolute URL for the Write endpoint where to flush the time series.
-	URL null.String `json:"url" envconfig:"K6_PROMETHEUS_REMOTE_URL"`
+	URL null.String `json:"url" envconfig:"K6_PROMETHEUS_RW_REMOTE_URL"`
```
na--, line 28: hmm this seems like duplication, how about just `K6_PROMETHEUS_RW_URL`? or `K6_PROMETHEUS_RW_SERVER_URL` / `K6_PROMETHEUS_RW_SERVER_ADDR`

pkg/remotewrite/config.go
```diff
@@ -194,49 +194,49 @@ func parseEnvs(env map[string]string) (Config, error) {
 	}
 
 	// envconfig is not processing some undefined vars (at least duration) so apply them manually
```
na--, line 196: not related to this PR, but I think this comment and all of the `envconfig:` struct tags are not needed, it seems like the extension is not using the `envconfig` at all

pkg/remotewrite/config.go
codebien: I'm fine with `K6_PROMETHEUS_RW_URL` and `K6_PROMETHEUS_RW_SERVER_URL`. `K6_PROMETHEUS_RW_SERVER_ADDR` isn't good because often the implementations require a specific path, so it has to be passed here also.

For example this is the default value:

https://github.com/grafana/xk6-output-prometheus-remote/blob/203e3b6105a3f5c679dd5a880fdbb8523cbe046f/pkg/remotewrite/config.go#L18

na--: `K6_PROMETHEUS_RW_SERVER_URL` has my vote then, though I am also fine with just `K6_PROMETHEUS_RW_URL`

pkg/remotewrite/config.go
codebien: The attempt was to refactor for polishing the code in the next days because I didn't plan this PR. Anyway, now we have this PR, so it should be fine to address it directly here.

register.go
```diff
@@ -6,7 +6,7 @@ import (
 )
 
 func init() {
-	output.RegisterExtension("output-prometheus-remote", func(p output.Params) (output.Output, error) {
+	output.RegisterExtension("xk6-prometheus-rw", func(p output.Params) (output.Output, error) {
```
imiric, line 9: We don't want to use `experimental-*` here?

And does `xk6-*` make sense, considering this extension will be imported directly by k6, without using xk6?

imiric (approved):
LGTM, besides the prefix question, and that I'm not sure about the envconfig removal, but @na-- would know better.

Also, could we add some E2E tests to this repo that actually test against Prometheus? I can open an issue if you agree.

na-- (approved):

register.go
na--: We can't use `experimental-`, since that will already be built into k6 and cause an error. 

In other extensions, we kept the xk6 module registration even after we bundled them into the core as `experimental` because some users probably already had a setup that used them as xk6 extensions. In this case, since we are changing the extension name and all of its parameters, it probably males less sense, so I am fine either way - leave it as it was, change it, doesn't really matter all that much :sweat_smile: 

Though some xk6 registration _should_ remain, since this is still going to be an xk6 extension in its own repo, even after it is bundled in `experimental`. We will make changes here and only occasionally merge them back in k6, so some user might want to use the latest and greatest version of the extension through xk6.

PR #83, opened 2022-12-13
Add k6's linting rules and address detected issues

oleiade:
Here's a PR adding the k6 golangci-lint configuration to the repo and addressing the issues outlined by it. Note that in a bunch of places, I simply locally deactivated the lining rule as I was not 100% certain if things such as type casts being unchecked were intentional or not.

Let me know what you think, and if you spot things that could/should be done differently 🙇🏻

### DIFF
--- a/pkg/remote/client.go
+++ b/pkg/remote/client.go
@@ -1,3 +1,4 @@
+// Package remote implements the Prometheus remote write protocol.
 package remote
 
 import (
@@ -15,6 +16,7 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+// HTTPConfig holds the config for the HTTP client.
 type HTTPConfig struct {
 	Timeout   time.Duration
 	TLSConfig *tls.Config
@@ -22,6 +24,7 @@ type HTTPConfig struct {
 	Headers   http.Header
 }
 
+// BasicAuth holds the config for basic authentication.
 type BasicAuth struct {
 	Username, Password string
 }
@@ -35,6 +38,7 @@ type WriteClient struct {
 	cfg *HTTPConfig
 }
 
+// NewWriteClient creates a new WriteClient.
 func NewWriteClient(endpoint string, cfg *HTTPConfig) (*WriteClient, error) {
 	if cfg == nil {
 		cfg = &HTTPConfig{}
@@ -89,8 +93,17 @@ func (c *WriteClient) Store(ctx context.Context, series []*prompb.TimeSeries) er
 	if err != nil {
 		return fmt.Errorf("HTTP POST request failed: %w", err)
 	}
-	defer resp.Body.Close()
-	io.Copy(io.Discard, resp.Body) //nolint:errcheck
+	defer func() {
+		err = resp.Body.Close()
+		if err != nil {
+			panic(err)
+		}
+	}()
+
+	_, err = io.Copy(io.Discard, resp.Body)
+	if err != nil {
+		return err
+	}
 
 	return validateResponseStatus(resp.StatusCode)
 }
--- a/pkg/remotewrite/config.go
+++ b/pkg/remotewrite/config.go
@@ -21,8 +21,10 @@ const (
 	defaultMetricPrefix = "k6_"
 )
 
+//nolint:gochecknoglobals
 var defaultTrendStats = []string{"p(99)"}
 
+// Config contains the configuration for the Output.
 type Config struct {
 	// ServerURL contains the absolute ServerURL for the Write endpoint where to flush the time series.
 	ServerURL null.String `json:"url"`
@@ -81,7 +83,7 @@ func (conf Config) RemoteConfig() (*remote.HTTPConfig, error) {
 	}
 
 	hc.TLSConfig = &tls.Config{
-		InsecureSkipVerify: conf.InsecureSkipTLSVerify.Bool,
+		InsecureSkipVerify: conf.InsecureSkipTLSVerify.Bool, //nolint:gosec
 	}
 
 	if len(conf.Headers) > 0 {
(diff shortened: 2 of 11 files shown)


pkg/remotewrite/remotewrite.go
```diff
@@ -138,15 +145,16 @@ func (o *Output) staleMarkers(t time.Time) []*prompb.TimeSeries {
 //
 // TODO: refactor, the code can be improved
 func (o *Output) setTrendStatsResolver(trendStats []string) error {
-	var trendStatsCopy []string
+	trendStatsCopy := make([]string, len(trendStats))
```
oleiade, line 148: This is wrong, I blindly listened to the linter 🤦🏻

oleiade: ✅

oleiade:
Note that this test: https://github.com/grafana/xk6-output-prometheus-remote/pull/83/files#diff-648d645d09ad7c92e6684cc66ecc0ca1a7ef72ee2f3bc1bf941aa5349357bba7R112 does not support parallel testing 👍🏻

pkg/remotewrite/prometheus_test.go
```diff
@@ -37,7 +37,7 @@ func TestMapSeries(t *testing.T) {
 }
 
 // buildTimeSeries creates a TimSeries with the given name, value and timestamp
-func buildTimeSeries(name string, value float64, timestamp time.Time) *prompb.TimeSeries {
+func buildTimeSeries(name string, value float64, timestamp time.Time) *prompb.TimeSeries { //nolint:unparam
```
codebien, line 40: What is the error here?

pkg/remote/client.go
```diff
@@ -1,3 +1,4 @@
+// Package remote implements the Prometheus remote write protocol.
```
codebien, line 1:
```suggestion
// Package remote implements a sender for the Prometheus remote write protocol.
```

pkg/remotewrite/config.go
```diff
@@ -21,8 +21,10 @@ const (
 	defaultMetricPrefix = "k6_"
 )
 
+//nolint:gochecknoglobals
```
codebien, line 24: mmm...should we make it not global? @oleiade what is your opinion? do you have a suggestion? I would avoid splitting it from the consts. Should we create a `defaultConfig` function?

pkg/remotewrite/remotewrite.go
```diff
@@ -138,7 +145,7 @@ func (o *Output) staleMarkers(t time.Time) []*prompb.TimeSeries {
 //
 // TODO: refactor, the code can be improved
 func (o *Output) setTrendStatsResolver(trendStats []string) error {
-	var trendStatsCopy []string
+	var trendStatsCopy []string //nolint:prealloc
```
codebien, line 148: Mmm..can we do something smarter to remove it? :thinking:

pkg/remotewrite/remotewrite.go
```diff
@@ -245,7 +252,7 @@ func (o *Output) convertToPbSeries(samplesContainers []metrics.SampleContainer)
 				swm.Latest = truncTime
 				o.tsdb[sample.TimeSeries] = swm
 				seen[sample.TimeSeries] = struct{}{}
-			} else {
+			} else { //nolint:gocritic
```
codebien, line 255: What is the error here? :thinking:

pkg/remotewrite/remotewrite.go
```diff
@@ -341,7 +349,13 @@ func (swm seriesWithMeasure) MapPrompb() []*prompb.TimeSeries {
 		newts = trend.MapPrompb(swm.TimeSeries, swm.Latest)
 
 	default:
-		panic(fmt.Sprintf("Something is really off, as I cannot recognize the type of metric %s: `%s`", swm.Metric.Name, swm.Metric.Type))
+		panic(
+			fmt.Sprintf(
+				"Something is really off, as I cannot recognize the type of metric %s: `%s`",
```
codebien, line 354: If you have suggestions for improving this message I'm happy, I don't like it.

pkg/remotewrite/remotewrite_test.go
```diff
@@ -109,6 +109,7 @@ func TestOutputConvertToPbSeries(t *testing.T) {
 	assert.Equal(t, exp, pbseries)
 }
 
+//nolint:paralleltest,tparallel
```
codebien, line 112: Autospot as a reminder, so I can fix it later

pkg/remotewrite/remotewrite.go
```diff
@@ -311,6 +318,7 @@ func (swm seriesWithMeasure) MapPrompb() []*prompb.TimeSeries {
 		}
 	}
 
+	//nolint:forcetypeassert
```
codebien, line 321: What is the error?

pkg/remotewrite/trend.go
```diff
@@ -11,6 +11,7 @@ import (
 	"go.k6.io/k6/metrics"
 )
 
+// TrendStatsResolver is a map of trend stats name and their relative resolver function
```
codebien, line 14: Autospot as a reminder, the comment is nice but I guess it is not required to export it

register.go
```diff
@@ -1,3 +1,4 @@
+//nolint:revive,stylecheck
```
codebien, line 1: What is the error?

pkg/remotewrite/prometheus_test.go
oleiade: The linter says the `value` parameter has always the same value and can be removed, as the codebase has some todos left, I didn't want to assume anything about the further usage of this function and deactivated the lining locally.

pkg/remotewrite/remotewrite.go
oleiade: I tried, but because we essentially are performing a filter operation here, we can't know the size beforehand. An alternative would be to preallocate a slice with length 0, and capacity set to the length of `trendStats`. What way, worst case scenario, no values are filtered out, and the underlying dynamic array has pre-set enough storage.

Thus: 
```go
trendStatsCopy := make([]string, 0, len(trendStats))
```

What do you think?

pkg/remotewrite/remotewrite.go
oleiade: It complains that the `else { if {` could be replaced with `else if {`. This won't be true anymore, once the TODOs are addressed. Thus I deactivated it locally.

pkg/remotewrite/remotewrite.go
oleiade: It complains that the type casts are not checked. Which is indeed the case. The method doesn't return an error, so better safe than sorry I chose to keep the current behavior (panic) in case of issue, and deactivate the linter check locally.

pkg/remotewrite/remotewrite.go
codebien: It sounds fine to me

pkg/remotewrite/config.go
codebien: Let's keep it as it is and I will try to fix in my next refactor

register.go
oleiade: It requires that the package is documented. But the register code is essentially plumbing, and I didn't judge it necessary to document it. Let me know if you disagree 👍🏻

pkg/remotewrite/remotewrite.go
codebien: Can you add it as a comment to the nolint, please? I will look into it

pkg/remotewrite/remotewrite.go
codebien: It sounds good to me, can you add it, please?

register.go
codebien: I think it's difficult to remember, so we should add a comment for it, but at that point, we can directly comment with an obvious sentence and remove the nolint.

pkg/remote/client.go
```diff
@@ -89,8 +93,17 @@ func (c *WriteClient) Store(ctx context.Context, series []*prompb.TimeSeries) er
 	if err != nil {
 		return fmt.Errorf("HTTP POST request failed: %w", err)
 	}
-	defer resp.Body.Close()
-	io.Copy(io.Discard, resp.Body) //nolint:errcheck
+	defer func() {
+		err = resp.Body.Close()
+		if err != nil {
+			panic(err)
+		}
```
codebien, line 100: It is too aggressive here, we have to expect errors here so we need to manage that better. I will do it directly.

pkg/remote/client.go
```diff
@@ -89,8 +93,17 @@ func (c *WriteClient) Store(ctx context.Context, series []*prompb.TimeSeries) er
...
+	}()
+
+	_, err = io.Copy(io.Discard, resp.Body)
+	if err != nil {
+		return err
+	}
```
codebien, line 106: It is better to wrap it. Set as a reminder.

pkg/remotewrite/config.go
codebien: I have unresolved this as a reminder.

codebien (approved):
LGTM

PR #85, opened 2022-12-15
Stale marker accuracy

codebien:
~~It introduces a new option `K6_PROMETHEUS_RW_STALE_MARKERS=<bool>` for enabling the Staleness marking process at the end of the test. It also disables the option by default.~~ Moved to https://github.com/grafana/xk6-output-prometheus-remote/pull/86

It adjusts the stale marker accuracy by adding slight padding.

### DIFF
--- a/pkg/remote/client_test.go
+++ b/pkg/remote/client_test.go
@@ -10,8 +10,9 @@ import (
 	"testing"
 	"time"
 
-	"github.com/golang/snappy"
 	"github.com/grafana/xk6-output-prometheus-remote/pkg/stale"
+
+	"github.com/golang/snappy"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	prompb "go.buf.build/grpc/go/prometheus/prometheus"
--- a/pkg/remotewrite/remotewrite.go
+++ b/pkg/remotewrite/remotewrite.go
@@ -25,6 +25,7 @@ type Output struct {
 
 	config             Config
 	logger             logrus.FieldLogger
+	now                func() time.Time
 	periodicFlusher    *output.PeriodicFlusher
 	tsdb               map[metrics.TimeSeries]*seriesWithMeasure
 	trendStatsResolver map[string]func(*metrics.TrendSink) float64
@@ -55,6 +56,10 @@ func New(params output.Params) (*Output, error) {
 	o := &Output{
 		client: wc,
 		config: config,
+		// TODO: consider to do this function millisecond-based
+		// so we don't need to truncate all the time we invoke it.
+		// Before we should analyze if in some cases is it useful to have it in ns.
+		now:    time.Now,
 		logger: logger,
 		tsdb:   make(map[metrics.TimeSeries]*seriesWithMeasure),
 	}
@@ -93,8 +98,7 @@ func (o *Output) Stop() error {
 	if !o.config.StaleMarkers.Bool {
 		return nil
 	}
-
-	staleMarkers := o.staleMarkers(time.Now())
+	staleMarkers := o.staleMarkers()
 	if len(staleMarkers) < 1 {
 		o.logger.Debug("No time series to mark as stale")
 		return nil
@@ -109,10 +113,18 @@ func (o *Output) Stop() error {
 }
 
 // staleMarkers maps all the seen time series with a stale marker.
-func (o *Output) staleMarkers(t time.Time) []*prompb.TimeSeries {
-	timestamp := t.UnixMilli()
-	staleMarkers := make([]*prompb.TimeSeries, 0, len(o.tsdb))
+func (o *Output) staleMarkers() []*prompb.TimeSeries {
+	// Add 1ms so in the extreme case that the time frame
+	// between the last and the next flush operation is under-millisecond,
+	// we can avoid the sample being seen as a duplicate,
+	// if we force it in the future.
+	// It is essential because if it overlaps, the remote write discards the last sample,
+	// so the stale marker and the metric will remain active for the next 5 min
+	// as the default logic without stale markers.
+	timestamp := o.now().
+		Truncate(time.Millisecond).Add(1 * time.Millisecond).UnixMilli()
 
+	staleMarkers := make([]*prompb.TimeSeries, 0, len(o.tsdb))
 	for _, swm := range o.tsdb {
 		series := swm.MapPrompb()
 		// series' length is expected to be equal to 1 for most of the cases
--- a/pkg/remotewrite/remotewrite_test.go
+++ b/pkg/remotewrite/remotewrite_test.go
@@ -312,7 +312,11 @@ func TestOutputStaleMarkers(t *testing.T) {
 		Tags:   registry.RootTagSet(),
 	}
 
-	o := Output{}
+	o := Output{
+		now: func() time.Time {
+			return time.Unix(1, 0)
+		},
+	}
 	err := o.setTrendStatsResolver([]string{"p(99)"})
 	require.NoError(t, err)
 	trendSink, err := newExtendedTrendSink(o.trendStatsResolver)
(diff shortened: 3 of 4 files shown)


pkg/remote/client_test.go
```diff
@@ -172,6 +175,40 @@ func TestNewWriteRequestBody(t *testing.T) {
...
 }
 
+func TestNewWriteRequestBodyWithStaleMarker(t *testing.T) {
+	t.Parallel()
+
+	ts := []*prompb.TimeSeries{
+		{
+			Labels: []*prompb.Label{{Name: "label1", Value: "val1"}},
+			Samples: []*prompb.Sample{{
+				Value:     math.Float64frombits(0x7ff0000000000002),
```
imiric, line 185: To avoid redefining this, maybe export the `remotewrite.staleNaN` var, or if an import cycle is a problem, put it in a common package?

pkg/remote/client_test.go
```diff
@@ -172,6 +175,40 @@ func TestNewWriteRequestBody(t *testing.T) {
...
+
+	var series prompb.WriteRequest
+	err = proto.Unmarshal(sb, &series)
+	require.NoError(t, err)
+	require.NotEmpty(t, series.Timeseries[0])
+	require.NotEmpty(t, series.Timeseries[0].Samples)
+
+	isStaleNaN := func(v float64) bool {
+		return math.Float64bits(v) == uint64(0x7ff0000000000002)
+	}
```
imiric, line 205: Same with this. It would be better as an exported function in some common package.

pkg/remote/client_test.go
```diff
@@ -172,6 +175,40 @@ func TestNewWriteRequestBody(t *testing.T) {
...
 
+func TestNewWriteRequestBodyWithStaleMarker(t *testing.T) {
+	t.Parallel()
+
+	ts := []*prompb.TimeSeries{
+		{
+			Labels: []*prompb.Label{{Name: "label1", Value: "val1"}},
+			Samples: []*prompb.Sample{{
+				Value:     math.Float64frombits(0x7ff0000000000002),
+				Timestamp: time.Now().UnixMilli(),
```
imiric, line 186: Maybe store this timestamp or use a fixed value, so that you can assert you got the same value back below, instead of the current `NotEmpty()` assertion?

pkg/remotewrite/remotewrite.go
```diff
@@ -99,7 +108,12 @@ func (o *Output) Stop() error {
 	defer o.logger.Debug("Output stopped")
 	o.periodicFlusher.Stop()
 
-	staleMarkers := o.staleMarkers(time.Now())
+	// Add 1ms so in the extreme case
+	// that the time frame between the last
+	// and the current flush operations is under-millisecond then
+	// we can avoid that the flush is seen as duplicate keeping the accuracy in the future.
+	timestamp := time.Now().Truncate(time.Millisecond).Add(1 * time.Millisecond)
+	staleMarkers := o.staleMarkers(timestamp)
```
imiric, line 121: I can't say I understand the need and possible side-effects of this, but you'd know better. :smile:

Would it be possible to add a test for it, though? I guess if we're dealing with sub-1ms variations it would be extremely flaky, but have you run into this issue in real world tests? A fix like this should probably be done and discussed in a separate PR.

pkg/remote/client_test.go
```diff
@@ -3,15 +3,18 @@ package remote
...
 	"time"
 
+	"github.com/golang/snappy"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	prompb "go.buf.build/grpc/go/prometheus/prometheus"
+	"google.golang.org/protobuf/proto"
 )
 
 func TestNewWrtiteClient(t *testing.T) {
```
imiric, line 20: Not relevant to this PR, but:

```suggestion
func TestNewWriteClient(t *testing.T) {
```

pkg/remotewrite/remotewrite_test.go
```diff
@@ -341,3 +345,84 @@ func TestOutputStaleMarkers(t *testing.T) {
 		assert.True(t, math.IsNaN(markers[i].Samples[0].Value), "it isn't a StaleNaN value")
 	}
 }
+
+func TestOutputStopWithStaleMarkesDisabled(t *testing.T) {
```
imiric, line 349:
```suggestion
func TestOutputStopWithStaleMarkersDisabled(t *testing.T) {
```

Though:
- The test is named `WithStaleMarkersDisabled`, yet the config sets `StaleMarkers: true`. :confused:
- And isn't this test a duplicate, considering you test both the `true` and `false` case in `TestOutputStopWithStaleMarkes` below? So I would just remove it.

pkg/remotewrite/remotewrite_test.go
```diff
@@ -341,3 +345,84 @@ func TestOutputStaleMarkers(t *testing.T) {
...
+	// TODO: it isn't optimal to maintain
+	// it should be more stable with a mock of the client
+	// and check if Store is invoked should be
+	// a more stable method.
+	entries := logHook.Drain()
+	require.Len(t, entries, 6)
+	assert.Contains(t, entries[4].Message, "No time series to mark as stale")
+}
+
+func TestOutputStopWithStaleMarkes(t *testing.T) {
```
imiric, line 381:
```suggestion
func TestOutputStopWithStaleMarkers(t *testing.T) {
```

pkg/remotewrite/config_test.go
```diff
@@ -449,3 +449,36 @@ func TestConfigTrendStats(t *testing.T) {
...
+func TestOptionStaleMarker(t *testing.T) {
+	t.Parallel()
+
+	cases := map[string]struct {
+		arg     string
+		env     map[string]string
+		jsonRaw json.RawMessage
+	}{
+		"JSON": {jsonRaw: json.RawMessage(`{"staleMarkers":true}`)},
+		"Env":  {env: map[string]string{"K6_PROMETHEUS_RW_STALE_MARKERS": "true"}},
```
imiric, line 462: Can you add a test case to check that `StaleMarkers` is `false` by default?

imiric (changes requested):
The new option LGTM, but I think some test issues should be addressed.

codebien:
Changes addressed in https://github.com/grafana/xk6-output-prometheus-remote/pull/86. I created a new PR to have a consistent branch name.

pkg/remotewrite/remotewrite.go
codebien: The concept is that the Output (same as k6 is nanosecond based) instead Prometheus is millisecond-based. So during the conversion we lose the accuracy.

>but have you run into this issue in real-world tests

By design it is possible, in the operation `periodFlusher.Stop` it sends "the last batch" then the code arrives here and closes with the stale marker, if this two ops are done in a single millisecond then we will hit a clash and the stale marker will be discarded. "The last batch" is still a remote call so I don't expect in the real world we will hit the issue.

pkg/remotewrite/remotewrite.go
```diff
@@ -109,10 +113,18 @@ func (o *Output) Stop() error {
 }
 
 // staleMarkers maps all the seen time series with a stale marker.
-func (o *Output) staleMarkers(t time.Time) []*prompb.TimeSeries {
-	timestamp := t.UnixMilli()
-	staleMarkers := make([]*prompb.TimeSeries, 0, len(o.tsdb))
+func (o *Output) staleMarkers() []*prompb.TimeSeries {
+	// Add 1ms so in the extreme case that the time frame
+	// between the last and the next flush operation is under-millisecond,
+	// because we can avoid the sample is seen as a duplicate,
```
codebien, line 119:
```suggestion
	// we can avoid the sample being seen as a duplicate,
```

pkg/remotewrite/remotewrite.go
```diff
@@ -109,10 +113,18 @@ func (o *Output) Stop() error {
...
+	// if we force it in the future.
+	// It is essential because if it overlaps, the remote write discards the last sample,
+	// so the Stale marker and the metric will remain active for the next 5 min
```
codebien, line 122:
```suggestion
	// so the stale marker and the metric will remain active for the next 5 min
```

imiric (approved):
I haven't tested this, but your explanation and test change makes sense. :+1:

PR #91, opened 2022-12-19
Remove labels with empty values

afajl:
Needed to remove empty labels to make this work with Thanos receive.


codebien:
Hi @afajl,
thanks for your contribution. :bow: 

k6 doesn't add empty tags (so empty labels can't be generated). They are allowed as a custom tag defined by the user, are you scripting in the test this tag addition? Can you explain your use case and why you're doing it, please?

codebien:
Hey @afajl, you're right, and in any case this is not allowed by definition from the spec. The Prometheus server does not enforce it and just removes it. We need to address this, thanks for reporting.

codebien:
Hey @afajl, can you try the latest version and confirm that it works for you, please?

afajl:
Yeah, that works fine. Thanks!

PR #112, opened 2023-03-09
Dockerfiles bump

jwcastillo:
the versions of grafana prometheus are updated, and the images of alpine and golang too


codebien (approved):
Thanks :bow:

PR #114, opened 2023-03-27
Add prometheus variable for datasource

7olstoy:
These values of datasource were hardcoded and does not use the variable.

### DIFF
--- a/grafana/dashboards/dashboard-tests.json
+++ b/grafana/dashboards/dashboard-tests.json
@@ -1,4 +1,47 @@
 {
+  "__inputs": [
+    {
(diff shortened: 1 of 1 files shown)


javaducky (changes requested):
Hi @7olstoy! Thank you for your update. Looks like you're missing the definition for the `DS_PROMETHEUS` variable.

image

7olstoy:
Hey @javaducky, thank you for your catch!
Added the variable and the same requires as already exist in another files.

javaducky (approved):
LGTM! Thank you for fix.

PR #132, opened 2023-07-10
Fix arm build

mem:
When trying to build for ARM (linux/arm), you get:

```
pkg/remote/client.go:120:35: cannot use 0xffffffff (untyped int constant 4294967295) as int value in argument to fmt.Errorf (overflows)
```

Pass the constant as unsigned integer.

### DIFF
--- a/pkg/remote/client.go
+++ b/pkg/remote/client.go
@@ -117,7 +117,7 @@ func newWriteRequestBody(series []*prompb.TimeSeries) ([]byte, error) {
 	}
 	if snappy.MaxEncodedLen(len(b)) < 0 {
 		return nil, fmt.Errorf("the protobuf message is too large to be handled by Snappy encoder; "+
-			"size: %d, limit: %d", len(b), 0xffffffff)
+			"size: %d, limit: %d", len(b), uint(0xffffffff))
 	}
 	return snappy.Encode(nil, b), nil
 }


pkg/remote/client.go
```diff
@@ -117,7 +117,7 @@ func newWriteRequestBody(series []*prompb.TimeSeries) ([]byte, error) {
 	}
 	if snappy.MaxEncodedLen(len(b)) < 0 {
 		return nil, fmt.Errorf("the protobuf message is too large to be handled by Snappy encoder; "+
-			"size: %d, limit: %d", len(b), 0xffffffff)
+			"size: %d, limit: %d", len(b), uint(0xffffffff))
```
codebien, line 120: It sounds to me the issue is more running on a 32-bit platform than being ARM. Am I missing something?

If my assumption is correct than doing the cast would generate a bad error message because the value is different based on the bit size.

Something like the following should be more accurate:

```
maxLen = math.MaxInt32
if bits.UintSize == 64 {
  maxLen = 0xffffffff
}
```

codebien (changes requested):

mstoykov:
Unfortunately k6 "dropped" 32 bit support a while back https://github.com/grafana/k6/issues/1999 mostly due to lack of anyone using it and there being problems we will likely find only through a lot of usage.

I feel like the problem here is that @mem you compiled for the "wrong" arm variant?

Also the main k6 repo currently outputs:
```
$ GOARCH=386 go build .
# github.com/grafana/xk6-output-prometheus-remote/pkg/remote
vendor/github.com/grafana/xk6-output-prometheus-remote/pkg/remote/client.go:120:35: cannot use 0xffffffff (untyped int constant 4294967295) as int value in argument to fmt.Errorf (overflows)
# go.k6.io/k6/output/cloud/expv2
output/cloud/expv2/metrics_client.go:89:35: cannot use 0xffffffff (untyped int constant 4294967295) as int value in argument to fmt.Errorf (overflows)
# github.com/grafana/xk6-webcrypto/webcrypto
vendor/github.com/grafana/xk6-webcrypto/webcrypto/aes.go:592:38: cannot use 549755813632 (untyped int constant) as int value in constant declaration (overflows)
```

pkg/remote/client.go
mem: ah, you are right, nice catch

mstoykov:
I will close this now. 

Please open an issue if this is still needed so we can dicuss it

PR #134, opened 2023-07-31
Migrate to Buf remote packages

### LINKED ISSUE #129: Migrate to Buf remote packages
Buf no longer supports the remote generation alpha (Go packages starting with `go.buf.build`). There are steps showing how to migrate to the new supported mechanism here: https://buf.build/docs/migration-guides/migrate-remote-generation-alpha/#go-module-proxy.

I'm happy to open a PR to migrate this project to use the new dependency. This will have the benefit of keeping up with future google protobuf-go releases in the future.

Reference: https://github.com/grafana/xk6-output-prometheus-remote/blob/34469a719af7149ae23c9fcdfedadabd847b18cf/go.mod#L12

codebien:
The Buf remote generation alpha is deprecated and projects should migrate to use remote packages. Update the protobuf dependencies to the remote package (and change the import path).

Fixes #129.


codebien:
#130 I pushed a rebased branch preserving the original author

mstoykov (approved):

PR #143, opened 2023-09-21
TLS certificates

codebien:


pkg/remotewrite/config.go
```diff
@@ -41,6 +41,12 @@ type Config struct {
 	// Password is the Password for the Basic Auth.
 	Password null.String `json:"password"`
 
+	// ClientCertificate is the public key of the SSL certificate.
```
olegbespalov, line 44: Here (and the same comment for key's key), I believe worth mentioning that the path is expected, not the data

olegbespalov (changes requested):

pkg/remotewrite/config.go
codebien: Done

pkg/remotewrite/config.go
```diff
@@ -89,6 +95,11 @@ func (conf Config) RemoteConfig() (*remote.HTTPConfig, error) {
 		InsecureSkipVerify: conf.InsecureSkipTLSVerify.Bool, //nolint:gosec
 	}
 
+	if conf.ClientCertificate.Valid && conf.ClientCertificateKey.Valid {
+		cert, _ := tls.LoadX509KeyPair(conf.ClientCertificate.String, conf.ClientCertificateKey.String)
```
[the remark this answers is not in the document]
codebien: Done. Added also a test for it

pkg/remotewrite/config.go
```diff
@@ -229,6 +255,14 @@ func parseEnvs(env map[string]string) (Config, error) {
 		c.Password = null.StringFrom(password)
 	}
 
+	if clientCertificate, certDefined := env["K6_PROMETHEUS_RW_CLIENT_CERTIFICATE"]; certDefined {
```
codebien, line 258:
```suggestion
	if clientCertificate, certDefined := env["K6_PROMETHEUS_RW_TLS_CLIENT_CERTIFICATE"]; certDefined {
```
Do you think we should be more specific about the naming?

olegbespalov: Both fine fro me :shrug:

pkg/remotewrite/config.go
```diff
@@ -41,6 +41,16 @@ type Config struct {
 	// Password is the Password for the Basic Auth.
 	Password null.String `json:"password"`
 
+	// ClientCertificate is the public key of the SSL certificate.
+	// It is expected the path of the certificate on the file system.
+	// If it is required a dedicated Certifacate Authority then it should be added
```
olegbespalov, line 46: It theory we could also provide the way to add the CA like it's implemented in k6, but not sure if it's under the scope of this PR

olegbespalov (approved):
:+1:

pkg/remotewrite/config.go
codebien: I would prefer if we get an explicit request from users for it. The demand for `mTLS` feature seems to be not so high in general.

pkg/remotewrite/config.go
mstoykov: Is there something else that we can have a client certificate for ?

If not I am also :shrug: - if yes , then I would prefer to have `mtls` even.

mstoykov (approved):
LGTM!

pkg/remotewrite/config.go
codebien: We are doing mostly the same with Username and Password where we aren't specifying that it is for BasicAuth.

Issue #101, opened 2023-02-07
Better API for HTTP Headers


codebien:
k6 does not support the most used case of the current API for setting the custom HTTP Headers, which is the Cortex authentication. See https://github.com/grafana/k6/pull/2904 for details.

The suggestion is to deprecate the current API and replace it with a fixed environment variable name supported by k6. (e.g. `K6_PROMETHEUES_RW_HTTP_HEADERS` and use the value for the key-values binding. k6 uses the same approach for the tags' customization in the tagsInfluxDBv1 output.

```
K6_PROMETHEUES_RW_HTTP_HEADERS=X-Scope-OrgID:<org>,X-ANOTHER_HEADER:val
```

lanore78:
Hi @codebien 

It would be really great if you let me know the estimated time for this issue to be released.
We've been researching new performance test tools that can replace LoadRunner, and currently K6 is the most likely tools to replace LoadRunner.

codebien:
Hi @lanore78,
>It would be really great if you let me know the estimated time for this issue to be released.

We don't have an estimated time for this, it could be considered in the next weeks. If you want to speed up the process, you may consider contributing with a PR. The required code should be doable for an external contributor with a bit of Go knowledge.

PR #147, opened 2023-09-22
New format for custom HTTP headers

codebien:
Closes https://github.com/grafana/xk6-output-prometheus-remote/issues/101

### DIFF
--- a/pkg/remotewrite/config.go
+++ b/pkg/remotewrite/config.go
@@ -180,7 +180,9 @@ func GetConsolidatedConfig(jsonRawConf json.RawMessage, env map[string]string, _
 }
 
 func parseEnvs(env map[string]string) (Config, error) {
-	var c Config
+	c := Config{
+		Headers: make(map[string]string),
+	}
 
 	getEnvBool := func(env map[string]string, name string) (null.Bool, error) {
 		if v, vDefined := env[name]; vDefined {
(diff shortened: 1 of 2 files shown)


pkg/remotewrite/config.go
```diff
@@ -237,6 +237,19 @@ func parseEnvs(env map[string]string) (Config, error) {
 		c.Headers[k] = v
 	}
 
+	if headers, headersDefined := env["K6_PROMETHEUS_RW_HTTP_HEADERS"]; headersDefined {
+		if c.Headers == nil {
+			c.Headers = make(map[string]string)
```
olegbespalov, line 242: How about moving this into the initialization to the top by default:

```diff
-var c Config
+c := Config{Headers: make(map[string]string)}
```

There is no significant difference in efficiency, but the code's reader will be a bit happier without 
`if c.Headers == nil`

pkg/remotewrite/config.go
```diff
@@ -237,6 +237,19 @@ func parseEnvs(env map[string]string) (Config, error) {
...
+		}
+		for _, kvPair := range strings.Split(headers, ",") {
+			header := strings.Split(kvPair, ":")
+			if len(header) != 2 {
+				continue
```
olegbespalov, line 247: Maybe we could log a warning that something is ignored (this `contunue` happening), it would be better for the UX

olegbespalov (commented):
Looks good, left a two minor discussable comments

pkg/remotewrite/config.go
codebien: I thought we could just return an error and avoid adding the logger dependency.

olegbespalov (approved):
:+1:

pkg/remotewrite/config.go
olegbespalov: Yeah, also works since it attracts the user attention :+1:

mstoykov (approved):

PR #148, opened 2023-09-22
Update Dashboards and Docker Compose setup

ppcano:
Update the existing test result dashboard, it is now called `k6 Prometheus`.

It also creates a new similar dashboard, `k6 Prometheus - Native Histograms`, to be used when storing trend metrics in native histograms.

These dashboards are an advancement from the existing one and #142. A big thanks to @jwcastillo for their exceptional work and contributions.

Docker compose changes
- Update Grafana and Prometheus versions
- Remove k6 service from the docker compose setup. This extension is built in core, so user can follow [docs instructions](https://k6.io/docs/results-output/real-time/prometheus-remote-write/#docker-compose-example). 

Other changes
- Remove examples folder
- Remove dashboards: Apdex, custom metrics. 
- Remove the test list dashboard. You can now select tests from a dropdown menu (dynamic variable).

### k6 Prometheus - Native Histograms

Dynamic variables to configure the dashboard queries:
- Change Prometheus datasource
- Filter TestID when `testid` tag is present
- Quantile calculation for Trend metrics
- AnalyzeMetric: select a k6 metric to analyze on the panel at the bottom
- AdhocFilter to filter all the mtrics

Screenshot 2023-09-22 at 20 14 57

Screenshot 2023-09-22 at 20 15 07

Screenshot 2023-09-22 at 20 15 15

Screenshot 2023-09-22 at 20 15 24

Screenshot 2023-09-22 at 20 15 34

### k6 Prometheus

This dashboard is similar than the previous one. Instead of `Quantile`, it has `Trend Metrics Query` to select the Stat depending on the values from the `K6_PROMETHEUS_RW_TREND_STATS ` option. 

Screenshot 2023-09-25 at 11 01 23


docker-compose.yml
```diff
@@ -1,26 +1,24 @@
 version: '3.8'
 
 networks:
-  k6:
   grafana:
   prometheus:
```
codebien, line 5:
```suggestion
  k6:
```
We should do the opposite, having a unique k6 network.

grafana/dashboards/k6-prometheus-native-histogram.json
codebien, line 1: 1. When I open the dashboard the first time, it should keep open the Performance Overview section. It looks better to me. 
2. Why the default quantile is 99? Why do we call it here quantile and p9x in the other? We should pick only one name to have a consistent glossary.
3. The `AnalyzeMetric` on top creates confusion, it applies only to one chart and not to all of them as you expect from something on top. Also, in the end, having a panel for doing what the Explore functionality in Grafana already does seems not optimal to me. For these two reasons, I'm for us dropping completely the Analyze panel.
4. I think HTTP section should stay before Checks section.
5. I don't know if there is a way to do it, but if I select a `testID` then I would like to get the time range that centers the test and not use a relative time range.
6. The Adhoc filters adds confusion, it pulls series not relevant for the k6 context. We should drop it, because if the user needs to add custom queries then the Explorer is the correct tool.

grafana/dashboards/k6-prometheus-native-histogram.json
```diff
@@ -0,0 +1,2125 @@
...
+          "value": [
+            "$__all"
+          ]
+        },
+        "datasource": {
+          "type": "prometheus",
+          "uid": "${DS_PROMETHEUS}"
+        },
+        "definition": "label_values(testid)",
+        "description": "Filter by \"testid\" tag (k6 run --tag testid=xyz)",
```
codebien, line 2034:
```suggestion
        "description": "Filter by \"testid\" tag, define it by tagging (k6 run --tag testid=xyz)",
```

codebien (commented):
Thanks for the effort @ppcano! :bow: 

I added some comments.

grafana/dashboards/k6-prometheus-native-histogram.json
ppcano: https://github.com/grafana/xk6-output-prometheus-remote/pull/148/commits/9ddf32d82e49b595cf3215021e64f17100c3cec2

docker-compose.yml
ppcano: https://github.com/grafana/xk6-output-prometheus-remote/pull/148/commits/7d7c2ea62c971cd2af74db5f69061f10fa96d2b9

ppcano:
@codebien 

> When I open the dashboard the first time, it should keep open the Performance Overview section. It looks better to me.

✅ All the sections are open now. 


> Why the default quantile is 99? Why do we call it here quantile and p9x in the other? We should pick only one name to have a consistent glossary.

The format of both options cannot be the same.

1. For native histograms, it uses the  [histogram_quantile() function](https://prometheus.io/docs/prometheus/latest/querying/functions/#histogram_quantile), requiring a value between 0 and 1. 

2. Without native histograms, the available options depends on the values of [K6_PROMETHEUS_RW_TREND_STATS](https://k6.io/docs/results-output/real-time/prometheus-remote-write/#counter-and-gauges).  Stats functions: count, sum, min, max, avg, med, p(x).


Although the situation is not ideal, exposing this info to the end-user could guide them to learn how this output and the k6 metric format work. 

The default for NH is 0.99 because is the default of `K6_PROMETHEUS_RW_TREND_STATS`.


For consistency, I have now renamed the variable name.



Screenshot 2023-09-26 at 11 40 33

Screenshot 2023-09-26 at 12 41 56

<p>&nbsp;</p>

> The AnalyzeMetric on top creates confusion, it applies only to one chart and not to all of them as you expect from something on top. Also, in the end, having a panel for doing what the Explore functionality in Grafana already does seems not optimal to me. For these two reasons, I'm for us dropping completely the Analyze panel.

✅ ok - I agree is redundant and adds an unnecessary overhead.

My intention here was to "guide" users how to query k6 metrics. Wdyt about adding an informative panel like:

Screenshot 2023-09-26 at 12 49 32

<p>&nbsp;</p>

> I think HTTP section should stay before Checks section.

✅ 


> I don't know if there is a way to do it, but if I select a testID then I would like to get the time range that centers the test and not use a relative time range.

This option is not available. 

<p>&nbsp;</p>

> The Adhoc filters adds confusion, it pulls series not relevant for the k6 context. We should drop it, because if the user needs to add custom queries then the Explorer is the correct tool.

Agree, it adds confusion. To improve this, I have opened a [feature request in the grafana repo](https://github.com/grafana/grafana/issues/75402).

I'd preserve the AdhocFilter because:

1. The majority of the dashboard users might not know how k6 metrics and tags(labels) work, so this menu would give visibility that they can filter results by k6 tags. It is a hook to be more proactive and start querying k6 metrics, not only visualizing the current panels. 

2. Without this option, filters could not be applied to the existing visualizations.

docker-compose.yml
```diff
@@ -2,48 +2,28 @@ version: '3.8'
...
-    image: grafana/grafana:10.0.1
+    image: grafana/grafana:10.1.2
     networks:
-      - grafana
-      - prometheus
+      - k6
     ports:
       - "3000:3000"
     environment:
       - GF_AUTH_ANONYMOUS_ORG_ROLE=Admin
```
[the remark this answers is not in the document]
ppcano: I guess this is the reason of why the docker compose setup on your k6-dashboard PR is different. 

This is weird.  This does not happen to me.


Screenshot 2023-09-26 at 15 56 38

Screenshot 2023-09-26 at 15 56 46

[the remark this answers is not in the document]
ppcano: Let's review this in a following PRs.

PR #151, opened 2023-09-28
Update readme: usage and development sections

ppcano:


README.md
```diff
@@ -1,30 +1,29 @@
...
 
-Then run new k6 binary with the following command for using the default configuration (e.g. remote write server url set to `http://localhost:9090/api/v1/write`):
-```
-./k6 run -o xk6-prometheus-rw script.js 
+### Development
+
+For developing or testing this extension, you can build a k6 binary with the local extension using [xk6](https://github.com/grafana/xk6) with the following steps:
+
+```bash
+cd xk6-output-prometheus-remote
```
codebien, line 19:
```suggestion
```
It seems pretty obvious

README.md
```diff
@@ -1,30 +1,29 @@
 # xk6-output-prometheus-remote
 
-[k6](https://github.com/grafana/k6) extension for publishing test-run metrics to Prometheus via Remote Write endpoint.
+This extension allows publishing test-run metrics to Prometheus via Remote Write endpoint.
```
codebien, line 3:
```suggestion
Output extension that allows publishing test-run metrics to Prometheus via Remote Write endpoint.
```

I don't like starting with `this` but maybe @heitortsergent has a better suggestion,

oleiade (approved):

README.md
```diff
@@ -1,30 +1,28 @@
...
-Then run new k6 binary with the following command for using the default configuration (e.g. remote write server url set to `http://localhost:9090/api/v1/write`):
-```
-./k6 run -o xk6-prometheus-rw script.js 
+### Development
+
+For developing or testing this extension, you can build a k6 binary with the local extension using [xk6](https://github.com/grafana/xk6) with the following steps:
+
+```bash
+
+xk6 build --with github.com/grafana/xk6-output-prometheus-remote=.
```
heitortsergent, line 20: Is this command building a k6 binary using the files from the main branch in this repo? If yes, would it make sense to update this to say something along the lines of "clone this repository, make changes, and run this command to build a new binary:", and then update the bash example?

README.md
heitortsergent: I think I like the original version better tbh... But an alternative if you don't wanna start with `this`, and I agree with that feedback:

"The xk6-output-prometheus-remote extension allows you to publish test-run metrics to Prometheus via Remote Write endpoint."
(repetition isn't always bad)

Another feedback, how about including a link to the [k6 extensions page](https://k6.io/docs/extensions/)? If a user lands in this repository and doesn't know much about k6 extensions, I think that might give them the best explanation of what this is.

heitortsergent (commented):
Just a couple of comments, but no blockers. :)

codebien (approved):

README.md
ppcano: https://github.com/grafana/xk6-output-prometheus-remote/pull/151/commits/21a830d8f73b1cc6f571e96226ed2cd9b9355bc1

README.md
ppcano: This is an advanced case, and requires a few more instructions and context. Then, we either document the entire process (which seems out of the scope) or refer to the primary guides.

In this case:
- shows the core command and try to highlight the simplicity
- refer to the docs

PR #158, opened 2023-10-31
Fix 32-bit build

mem:
When trying to build for 32-bit ARM (linux/arm), you get:

```
pkg/remote/client.go:120:35: cannot use 0xffffffff (untyped int constant 4294967295) as int value in argument to fmt.Errorf (overflows)
```

Pass the constant as unsigned 64-bit integer, which is what the underlying snappy code is doing (it's casting the argument to an uint64 and comparing against 0xffffffff, which in Go means it's comparing against uint64(0xffffffff)). Since this is just an error message, it shouldn't matter, but let's be consistent.

### DIFF
--- a/pkg/remote/client.go
+++ b/pkg/remote/client.go
@@ -117,7 +117,7 @@ func newWriteRequestBody(series []*prompb.TimeSeries) ([]byte, error) {
 	}
 	if snappy.MaxEncodedLen(len(b)) < 0 {
 		return nil, fmt.Errorf("the protobuf message is too large to be handled by Snappy encoder; "+
-			"size: %d, limit: %d", len(b), 0xffffffff)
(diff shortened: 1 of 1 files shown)


mem:
Sorry about recreating the PR instead of reopening the old one (#132). My bad.

I reworded the commit message to reflect the fact that this is not specific to ARM.

The reason why I would like to have this is because I would like to run k6 on linux/arm (Raspberry Pi and whatnot).

codebien (changes requested):
Hey @mem,
I think my comment on the previous PR https://github.com/grafana/xk6-output-prometheus-remote/pull/132#discussion_r1259401503 is still valid, can you take a look into it, please?

codebien (approved):

mstoykov (approved):

Issue #93, opened 2022-12-27
Support sigv4 to push metric to AWS managed prometheus


boskiv:
Feature request: Add support for sigv4 method along with of username/password combinations to write to a remote AWS managed prometheus instance

https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write

obanby:
Should be addressed by https://github.com/grafana/xk6-output-prometheus-remote/pull/169

niroz89:
PR https://github.com/grafana/xk6-output-prometheus-remote/pull/169 covers the AWS Access Keys credential provider. It would be nice to cover other credential providers especially [Assume role credential provider](https://docs.aws.amazon.com/sdkref/latest/guide/feature-assume-role-credentials.html) , so we can avoid hardcoding AWS access keys/secret and generate temporary token instead when running k6-oss in AWS environment.

PR #161, opened 2024-02-02
feat: Add SigV4 support

rybbchao:
Add SigV4 support for #93.

Environment variables:

- `K6_PROMETHEUS_RW_SIGV4_AUTH`: Enable **SigV4** for **AWS Managed Prometheus**.
- `K6_PROMETHEUS_RW_SIGV4_REGION`: AWS region where the workspace is.
- `K6_PROMETHEUS_RW_SIGV4_ACCESS_KEY`: AWS access key.
- `K6_PROMETHEUS_RW_SIGV4_SECRET_KEY`: AWS secret key.
- `K6_PROMETHEUS_RW_SIGV4_PROFILE`: AWS profile to use.
- `K6_PROMETHEUS_RW_SIGV4_ROLE_ARN`: AWS role ARN to assume.

Note: Set `AWS_SDK_LOAD_CONFIG` to `true` for the shared config and credential files.


mstoykov:
Looking at teh amount of dependancies this pulls in ... I am concerned.

I did a quick check and this adds 7mb on top of k6 for this one authentication. +Every security alert we get about all of those dependancies.

GIven that this currently goes in k6 core I really would prefer not to merge this.

Some quick googling shows that there are ways to get the authenticaiton without pulling in the whole aws sdk and I would recommend we actually go down this route

https://gist.github.com/anandkunal/b67eb94454b77cfc2b50026989586cc0 

cc @codebien

codebien (changes requested):
I agree, @rybbchao would you be interested in trying the path suggested by @mstoykov?

codebien:
Hey @rybbchao,
I'm going to close this issue as it seems is not under active development. Feel free to re-open if you find the time to re-work the changes.

Thanks for your contribution :bow:

PR #165, opened 2024-04-29
Fix Peak RPS calculation, shared tooltip and fix VU display in the Performance Overview 

Sea-you:
- Add a fix to the Peak RPS calculation by adding the Reduce (to max) transformation.
- Set shared tooltip in the Performance Overview panel
- Calculate the number of VUs by summing instead of average as it might mislead if you're testing with multiple instances. In our case 4 nodes, each 250 VUs. Previously this would have been displayed as 250 instead of 1000.

image

### DIFF
--- a/grafana/dashboards/k6-prometheus.json
+++ b/grafana/dashboards/k6-prometheus.json
@@ -20,7 +20,6 @@
   "fiscalYearStartMonth": 0,
   "gnetId": 19665,
   "graphTooltip": 2,
-  "id": 29255,
   "links": [
     {
       "asDropdown": false,
--- a/grafana/dashboards/k6-prometheus.json
+++ b/grafana/dashboards/k6-prometheus.json
@@ -40,7 +40,7 @@
     {
       "datasource": {
         "type": "prometheus",
-        "uid": "pZRohFJIk"
+        "uid": "${DS_PROMETHEUS}"
       },
       "fieldConfig": {
         "defaults": {
@@ -223,7 +223,7 @@
         {
           "datasource": {
             "type": "prometheus",
-            "uid": "pZRohFJIk"
+            "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
           "expr": "avg(k6_vus{testid=~\"$testid\"})",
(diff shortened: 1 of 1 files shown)


grafana/dashboards/k6-prometheus.json
```diff
@@ -18,7 +18,9 @@
   "description": "Visualize k6 OSS results stored in Prometheus",
   "editable": true,
   "fiscalYearStartMonth": 0,
-  "graphTooltip": 0,
+  "gnetId": 19665,
+  "graphTooltip": 2,
+  "id": 29255,
```
ppcano, line 23: `id` is specific to the running Grafana instance. It should be removed from the JSON configuration.

ppcano (changes requested):
@Sea-you, thanks for fixing this and the improvements. 

 I think it is ready after removing the `id` property.

grafana/dashboards/k6-prometheus.json
Sea-you: fixed!

grafana/dashboards/k6-prometheus.json
```diff
@@ -2014,13 +2042,13 @@
     ]
   },
   "time": {
-    "from": "now-5m",
```
ppcano, line 2017: @Sea-you please, do not change this option.

grafana/dashboards/k6-prometheus.json
```diff
@@ -2014,13 +2042,13 @@
...
+    "from": "now-6h",
     "to": "now"
   },
   "timepicker": {},
   "timezone": "",
   "title": "k6 Prometheus",
   "uid": "ccbb2351-2ae2-462f-ae0e-f2c893ad1028",
-  "version": 1,
+  "version": 6,
```
ppcano, line 2052: @Sea-you, I don't think we need to change this, but nvm, can we set this to version `2` instead?

Sea-you: fixed

grafana/dashboards/k6-prometheus.json
Sea-you: fixed

ppcano (approved):

ppcano:
@Sea-you, thanks for fixing the issue, improving the dashboard, and sharing it with the community!

PR #169, opened 2024-08-30
feat: support sigv4 signing

obanby:
- Added Tripper to enable signing requests
- Added Signer to implement the AWS sigv4 signing standard

I based my PR on this comment https://github.com/grafana/xk6-output-prometheus-remote/pull/161#issuecomment-1934440630 and  https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html

Note: I don't have deep knowledge about the sigv4, so please provide as much feedback as possible, and I will be happy to address it :) 

Addresses: https://github.com/grafana/xk6-output-prometheus-remote/issues/93

### DIFF
--- a/pkg/remotewrite/config.go
+++ b/pkg/remotewrite/config.go
@@ -3,6 +3,7 @@ package remotewrite
 import (
 	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"github.com/grafana/xk6-output-prometheus-remote/pkg/sigv4"
 	"net/http"
@@ -121,6 +122,14 @@ func (conf Config) RemoteConfig() (*remote.HTTPConfig, error) {
 			return nil, fmt.Errorf("failed to load the TLS certificate: %w", err)
 		}
 		hc.TLSConfig.Certificates = []tls.Certificate{cert}
+	}
+
+	if isSigV4PartiallyConfigured(conf.SigV4Region, conf.SigV4AccessKey, conf.SigV4SecretKey) {
+		return nil, errors.New(
+			"sigv4 seems to be partially configured. All of " +
+				"K6_PROMETHEUS_RW_SIGV4_REGION, K6_PROMETHEUS_RW_SIGV4_ACCESS_KEY, K6_PROMETHEUS_RW_SIGV4_SECRET_KEY " +
+				"must all be set. Unset all to bypass sigv4",
+		)
 	}
 
 	if conf.SigV4Region.Valid && conf.SigV4AccessKey.Valid && conf.SigV4SecretKey.Valid {
@@ -428,4 +437,13 @@ func parseArg(text string) (Config, error) {
 	}
 
 	return c, nil
+}
+
+func isSigV4PartiallyConfigured(region, accessKey, secretKey null.String) bool {
+	hasRegion := region.Valid && len(strings.TrimSpace(region.String)) != 0
+	hasAccessID := accessKey.Valid && len(strings.TrimSpace(accessKey.String)) != 0
+	hasSecretAccessKey := secretKey.Valid && len(strings.TrimSpace(secretKey.String)) != 0
+	// either they are all set, or all not set. False if partial
+	isComplete := (hasRegion && hasAccessID && hasSecretAccessKey) || (!hasRegion && !hasAccessID && !hasSecretAccessKey)
+	return !isComplete
 }
--- a/pkg/remotewrite/config.go
+++ b/pkg/remotewrite/config.go
@@ -5,11 +5,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/grafana/xk6-output-prometheus-remote/pkg/sigv4"
 	"net/http"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/grafana/xk6-output-prometheus-remote/pkg/sigv4"
 
 	"github.com/grafana/xk6-output-prometheus-remote/pkg/remote"
 	"go.k6.io/k6/lib/types"
--- a/pkg/remotewrite/config.go
+++ b/pkg/remotewrite/config.go
@@ -4,6 +4,7 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"github.com/grafana/xk6-output-prometheus-remote/pkg/sigv4"
 	"net/http"
 	"strconv"
 	"strings"
@@ -68,6 +69,15 @@ type Config struct {
 	TrendStats []string `json:"trendStats"`
 
 	StaleMarkers null.Bool `json:"staleMarkers"`
+
+	// Sigv4Region is the AWS region where the workspace is.
+	Sigv4Region null.String `json:"sigV4Region"`
+
+	// Sigv4AccessKey is the AWS access key.
+	Sigv4AccessKey null.String `json:"sigV4AccessKey"`
+
+	// Sigv4SecretKey is the AWS secret key.
+	Sigv4SecretKey null.String `json:"sigV4SecretKey"`
 }
 
 // NewConfig creates an Output's configuration.
@@ -81,6 +91,9 @@ func NewConfig() Config {
 		Headers:               make(map[string]string),
 		TrendStats:            defaultTrendStats,
 		StaleMarkers:          null.BoolFrom(false),
+		Sigv4Region:           null.NewString("", false),
+		Sigv4AccessKey:        null.NewString("", false),
+		Sigv4SecretKey:        null.NewString("", false),
 	}
 }
 
@@ -108,6 +121,14 @@ func (conf Config) RemoteConfig() (*remote.HTTPConfig, error) {
 			return nil, fmt.Errorf("failed to load the TLS certificate: %w", err)
 		}
 		hc.TLSConfig.Certificates = []tls.Certificate{cert}
+	}
+
+	if conf.Sigv4Region.Valid && conf.Sigv4AccessKey.Valid && conf.Sigv4SecretKey.Valid {
+		hc.SigV4 = &sigv4.Config{
+			Region:             conf.Sigv4Region.String,
+			AwsAccessKeyID:     conf.Sigv4AccessKey.String,
+			AwsSecretAccessKey: conf.Sigv4SecretKey.String,
+		}
 	}
 
 	if len(conf.Headers) > 0 {
@@ -147,6 +168,18 @@ func (conf Config) Apply(applied Config) Config {
 
 	if applied.BearerToken.Valid {
 		conf.BearerToken = applied.BearerToken
+	}
+
+	if applied.Sigv4Region.Valid {
+		conf.Sigv4Region = applied.Sigv4Region
+	}
+
+	if applied.Sigv4AccessKey.Valid {
+		conf.Sigv4AccessKey = applied.Sigv4AccessKey
+	}
+
+	if applied.Sigv4SecretKey.Valid {
+		conf.Sigv4SecretKey = applied.Sigv4SecretKey
 	}
 
 	if applied.PushInterval.Valid {
@@ -297,6 +330,18 @@ func parseEnvs(env map[string]string) (Config, error) {
 			}
 			c.Headers[header[0]] = header[1]
 		}
+	}
+
+	if sigv4Region, sigv4RegionDefined := env["K6_PROMETHEUS_RW_SIGV4_REGION"]; sigv4RegionDefined {
+		c.Sigv4Region = null.StringFrom(sigv4Region)
+	}
+
+	if sigv4AccessKey, sigv4AccessKeyDefined := env["K6_PROMETHEUS_RW_SIGV4_ACCESS_KEY"]; sigv4AccessKeyDefined {
+		c.Sigv4AccessKey = null.StringFrom(sigv4AccessKey)
+	}
+
+	if sigv4SecretKey, sigv4SecretKeyDefined := env["K6_PROMETHEUS_RW_SIGV4_SECRET_KEY"]; sigv4SecretKeyDefined {
+		c.Sigv4SecretKey = null.StringFrom(sigv4SecretKey)
 	}
 
 	if b, err := envBool(env, "K6_PROMETHEUS_RW_TREND_AS_NATIVE_HISTOGRAM"); err != nil {
--- a/pkg/remotewrite/config.go
+++ b/pkg/remotewrite/config.go
@@ -70,14 +70,14 @@ type Config struct {
 
 	StaleMarkers null.Bool `json:"staleMarkers"`
 
-	// Sigv4Region is the AWS region where the workspace is.
-	Sigv4Region null.String `json:"sigV4Region"`
+	// SigV4Region is the AWS region where the workspace is.
+	SigV4Region null.String `json:"sigV4Region"`
 
-	// Sigv4AccessKey is the AWS access key.
-	Sigv4AccessKey null.String `json:"sigV4AccessKey"`
+	// SigV4AccessKey is the AWS access key.
+	SigV4AccessKey null.String `json:"sigV4AccessKey"`
 
-	// Sigv4SecretKey is the AWS secret key.
-	Sigv4SecretKey null.String `json:"sigV4SecretKey"`
+	// SigV4SecretKey is the AWS secret key.
+	SigV4SecretKey null.String `json:"sigV4SecretKey"`
 }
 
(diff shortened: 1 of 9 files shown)


mike-sol:
This is great, enables easy export to Amazon Managed Prometheus. 👍

mike-sol:
Can I buy someone a coffee / beer / etc to get this reviewed and merged?

olegbespalov:
Hey @mike-sol !

The pull request is on our radar, but the team is busy with the upcoming K6 release. We'll return to it once we finish. Sorry for the delay!

pkg/sigv4/sigv4.go
```diff
@@ -0,0 +1,200 @@
...
+	)
+
+	payloadHash, err := d.getPayloadHash(req)
+	if err != nil {
+		return err
+	}
+
+	d.payloadHash = payloadHash
+	d.addRequiredHeaders(req)
+	d.canonicalHeaders, d.signedHeaders = d.getCanonicalAndSignedHeaders(req)
```
olegbespalov, line 56: Is there a big reason to keep the state (`canonicalHeaders` and `signedHeaders` and request related values) on the singer level instead of passing them downstream?

pkg/sigv4/sigv4.go
```diff
@@ -0,0 +1,200 @@
...
+	req.Header.Set("Host", req.Host)
+	req.Header.Set("x-amz-date", d.iSO8601Date)
+	req.Header.Set("x-amz-content-sha256", d.payloadHash)
+}
+
+func (d *DefaultSigner) getCanonicalAndSignedHeaders(req *http.Request) (string, string) {
+	var headers []string
+	var signedHeaders []string
+
+	for key, value := range req.Header {
```
olegbespalov, line 111: I'm checking the [AWS SDK implementation](https://github.com/aws/aws-sdk-go-v2/blob/main/aws/signer/v4/v4.go#L434-L436), and it has a bit more complicated logic, like skipping from signing some keys, do you think we need something similar here?

The same probably applies to the query params

pkg/sigv4/tripper_test.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"github.com/grafana/xk6-output-prometheus-remote/pkg/sigv4"
+	"github.com/stretchr/testify/assert"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestTripper_request_includes_required_headers(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
olegbespalov, line 12: Can we make this test `t.Parallel()` ?

pkg/sigv4/sigv4.go
```diff
@@ -0,0 +1,200 @@
...
+		d.signedHeaders,
+		signature,
+	)
+	req.Header.Set("Authorization", authorizationHeader)
+	return nil
+}
+
+func (d *DefaultSigner) getPayloadHash(req *http.Request) (string, error) {
+	if req.Body == nil {
+		return hex.EncodeToString(sha256.New().Sum(nil)), nil
```
olegbespalov, line 79: It could be just a [const](https://github.com/aws/aws-sdk-go-v2/blob/main/aws/signer/internal/v4/const.go#L6), like it's done here.

olegbespalov (commented):
Hi @obanby!

Thanks for the PR, and sorry for the delay in reviewing!

First of all, I'd like to say that, like you, I don't have a deep knowledge of sigv4. That's why I checked the AWS Go SDK to familiarize myself with it and made some of my comments based on the logic that I found implemented there.

Another question is, have you tested these changes if they work?

obanby:
> Hi @obanby!
> 
> Thanks for the PR, and sorry for the delay in reviewing!
> 
> First of all, I'd like to say that, like you, I don't have a deep knowledge of sigv4. That's why I checked the AWS Go SDK to familiarize myself with it and made some of my comments based on the logic that I found implemented there.
> 
> Another question is, have you tested these changes if they work?

Thanks @olegbespalov, yes I have tested with our aws managed prometheus.

pkg/sigv4/sigv4.go
```diff
@@ -0,0 +1,200 @@
...
+)
+
+const signingAlgo = "AWS4-HMAC-SHA256"
+const awsServiceName = "aps"
+
+type Signer interface {
+	Sign(req *http.Request) error
+}
+
+type DefaultSigner struct {
```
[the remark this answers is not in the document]
obanby: I can do that!

Just for my own knowledge, will maintainability of this be easier due to potential to reuse, or are there any other reasons?

pkg/sigv4/sigv4.go
obanby: addressed in [f3e0598](https://github.com/grafana/xk6-output-prometheus-remote/pull/169/commits/f3e0598fb8897a0600f8f5030b8696c25e1553d0)

pkg/sigv4/tripper_test.go
obanby: addressed in [f3e0598](https://github.com/grafana/xk6-output-prometheus-remote/pull/169/commits/f3e0598fb8897a0600f8f5030b8696c25e1553d0)

pkg/sigv4/sigv4.go
obanby: addressed in [f3e0598](https://github.com/grafana/xk6-output-prometheus-remote/pull/169/commits/f3e0598fb8897a0600f8f5030b8696c25e1553d0)

pkg/sigv4/sigv4.go
obanby: They were used in multiple places and I didn't want to pass them individually as arguments. 

I changed it as per your suggestion.

pkg/sigv4/tripper.go
```diff
@@ -0,0 +1,35 @@
...
+	next   http.RoundTripper
+}
+
+type Config struct {
+	Region             string
+	AwsSecretAccessKey string
+	AwsAccessKeyID     string
+}
+
+func NewRoundTripper(config *Config, next http.RoundTripper) *Tripper {
```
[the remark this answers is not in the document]
obanby: addressed in [f3e0598](https://github.com/grafana/xk6-output-prometheus-remote/pull/169/commits/f3e0598fb8897a0600f8f5030b8696c25e1553d0)

pkg/sigv4/utils.go
```diff
@@ -0,0 +1,98 @@
...
+		} else {
+			fmt.Fprintf(&buf, "%%%02X", c)
+		}
+	}
+	return buf.String()
+}
+
+// stripExcessSpaces will rewrite the passed in slice's string values to not
+// contain multiple side-by-side spaces.
+func stripExcessSpaces(str string) string {
```
olegbespalov, line 28: This method (and maybe others from this file) looks utterly identical to the one from AWS SDK. Unfortunately, I'm not sure if we can accept it as it is because of licensing. Its performance is also less of a concern here. Can we rewrite it (and other methods if needed) by hand?

pkg/sigv4/sigv4.go
```diff
@@ -0,0 +1,254 @@
...
+	config *Config
+}
+
+func NewDefaultSigner(config *Config) Signer {
+	// initialize noEscape array. This way we can avoid using init() functions
+	for i := 0; i < len(noEscape); i++ {
+		// AWS expects every character except these to be escaped
+		noEscape[i] = (i >= 'A' && i <= 'Z') ||
```
olegbespalov, line 30: If we populate this in a constructor, let's move this to a `Signer` property (since it's a state), and if needed, just pass downstream any method. The rationale is to avoid using a global state.

pkg/sigv4/utils.go
obanby: Yes these ones where a ported straight for the sdk. I can try to rewrite it by hand. I will look into the test cases for it too and see If I can get them to pass with the new rewrite.

obanby: There are couple more here that ported and made slight changes, would you like me to look into rewriting them?

olegbespalov: I mean, the rule of thumb is that we should not copy code directly, so yeah, if rewriting it on your own is preferable. But since, in any case, this PR is aiming to re-implement some logic to make the output AWS compatible, which still forces us that or another way use similar approaches so I believe at some point, we should still left some credits to AWS SDK with something like `logic is highly inspired by the AWS SDK` or any better referencing

obanby: Yes thats exactly it, I initially didn't want to rewrite it since I wanted the exact aws implementation. My thinking is we want to use the SDK behaviour, but without including it, so I figured porting it would be less risk of mistakes.

So are you okay if I add a comment with attributions instead of rolling a new one?

olegbespalov: No, that's not what I meant, sorry :cry: I was saying that in all cases, we need to mention that logic of that package tries to mimic AWS SDK. But having an utterly identical method, even if we reference to AWS SDK, isn't way to go

joanlopez (commented):
Hey @obanby, thanks for your contribution! 🙌🏻 

I have reviewed this PR as well, and I don't have any other comments beyond what @olegbespalov has already mentioned. So, feel free to try to address them all and I'll be happy to give my 👍🏻 once ready!

obanby:
Thanks @olegbespalov @joanlopez, I addressed all the comments. 

One thing I noticed, but didn't know how to go about it is: erroring out if only partial config is passed in. Say a user sets the the SIGV4_REGION without setting the access key. Any ideas how to fail with an informative message?

olegbespalov:
Hey @obanby 

Thanks for the update; I'll go through it at the beginning of next week!

> One thing I noticed, but didn't know how to go about it is: erroring out if only partial config is passed in. Say a user sets the the SIGV4_REGION without setting the access key. Any ideas how to fail with an informative message?

You could try collecting the validation errors in the slice, something like:

```go
errs := make([]string, 0)

if len(strings.TrimSpace(config.Region)) == 0 {
	errs = append(errs, "sigV4 config `Region` must be set")
}

if len(strings.TrimSpace(config.AwsSecretAccessKey)) == 0 {
	errs = append(errs, "sigV4 config `AwsSecretAccessKey` must be set")
}

if len(strings.TrimSpace(config.AwsAccessKeyID)) == 0 {
	errs = append(errs, "sigV4 config `AwsAccessKeyID` must be set")
}

if len(errs) > 0 {
	return nil, errors.New("can't initialize a sigv4 round tripper:" + strings.Join(errs, ", "))
}
```

obanby:
Thanks @olegbespalov, but to clarify. The current error handling should have caught that `config.AwsSecretAccessKey` is not set, and returns the error to the tripper client. However, that error doesn't correctly get propagated to console. So the user end up with a 403 unauthorized, because they didn't see any errors about sigv4. So my question is, how can this error propagate all the way up (preferably also short circuit the test)

olegbespalov:
Right, then I believe we should add validation somewhere around `RemoteConfig` e.g.  https://github.com/grafana/xk6-output-prometheus-remote/pull/169/files#diff-35f036a9d95de5edce2e31d56e65f748e31a64950ca6af842f241bf8dc01bf7fR126

```
sigvcfg := &sigv4.Config{
	Region:             conf.SigV4Region.String,
	AwsAccessKeyID:     conf.SigV4AccessKey.String,
	AwsSecretAccessKey: conf.SigV4SecretKey.String,
}

err := sigvcfg.Validate()
if err != nil {
    return nil, err
}

hc.SigV4 = sigvcfg
```

obanby:
Sounds good, I will try that. 

I will leave it till your review next week and address it then with any other comments you have :)

pkg/sigv4/sigv4.go
```diff
@@ -0,0 +1,244 @@
...
+	"io"
+	"net/http"
+	"net/url"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+)
+
+type Signer interface {
```
olegbespalov, line 18: If I'm not mistaken, we could make this interface (and some other structs in this package) unexported, which explicitly marks them for internal only (package) usage, helping with possible refactoring in the future and protecting other usage

olegbespalov:
Hey @obanby !

Just in case, I've reviewed the rest of the changes, and it looks good; maybe the only missing part is [the validation that we discussed before](https://github.com/grafana/xk6-output-prometheus-remote/pull/169#issuecomment-2407775184) to advise customers faster about misconfiguration.

So, thanks for implementing that!

obanby:
Thanks @olegbespalov was just taking sometime to see how to go about it. My last commit I addressed the partial commit issue, that's kinda what I was pointing out, see if you like that and we can go from there :)

pkg/remotewrite/config.go
```diff
@@ -110,6 +124,22 @@ func (conf Config) RemoteConfig() (*remote.HTTPConfig, error) {
...
+		)
+	}
+
+	if conf.SigV4Region.Valid && conf.SigV4AccessKey.Valid && conf.SigV4SecretKey.Valid {
+		hc.SigV4 = &sigv4.Config{
+			Region:             conf.SigV4Region.String,
+			AwsAccessKeyID:     conf.SigV4AccessKey.String,
+			AwsSecretAccessKey: conf.SigV4SecretKey.String,
+		}
+	}
```
olegbespalov, line 141: `isSigV4PartiallyConfigured` seems like a functional copy of `sigv4.Config.validate()`, it has a minor UX benefit that it explicitly refers to environment variables, but still, to be honest, it's better to have a single version so I have a minor suggestion to get rid of it and just make `sigv4.Config.validate()` exported and execute here, something like:

```suggestion
	if conf.SigV4Region.Valid && conf.SigV4AccessKey.Valid && conf.SigV4SecretKey.Valid {
		hc.SigV4 = &sigv4.Config{
			Region:             conf.SigV4Region.String,
			AwsAccessKeyID:     conf.SigV4AccessKey.String,
			AwsSecretAccessKey: conf.SigV4SecretKey.String,
		}
		
		if err := hc.SigV4.Validate(); err != nil {
		    return nil, fmt.Errorf(
			"sigv4 seems to be partially configured. All of " +
				"K6_PROMETHEUS_RW_SIGV4_REGION, K6_PROMETHEUS_RW_SIGV4_ACCESS_KEY, K6_PROMETHEUS_RW_SIGV4_SECRET_KEY " +
				"must all be set: %w.", err),
		}
	}
```

How does that sound?

obanby: The observation I have is that user can set `K6_PROMETHEUS_RW_SIGV4_ACCESS_KEY` and `K6_PROMETHEUS_RW_SIGV4_ACCESS_KEY` but forgot to set the region env variable. Now they will get 403 error but the error will be coming from upstream and won't be caught early since this condition will resolve to false ` conf.SigV4Region.Valid && conf.SigV4AccessKey.Valid && conf.SigV4SecretKey.Valid`

olegbespalov: Good point! :+1:

olegbespalov (approved):
Looks good to me! :rocket: Thanks for the contribution! :bow: 

Let's wait till @joanlopez will give a final look/feedback!

obanby:
I addressed some of the issues flagged by the CI lint job. 

@olegbespalov @joanlopez please have a look. I hope we can get this in soon :)

olegbespalov (approved):

obanby:
@olegbespalov @joanlopez, when will we be able to merge this please?

olegbespalov:
Hey @obanby, you have my approval, but we still need the review from @joanlopez for merging. 

In the meantime, could you please address comments from the linter? Most of them look like auto-fixes, so it's just a matter of running `golangci-lint run --fix` against the files. You could also validate if that helps by running the `make lint` command locally.

joanlopez (approved):

obanby:
Thanks all, I have address the changes flagged by `make lint`.

There are two missing, one is not related to my changes, and the other will require adding another function to breeak down the method, but doesn't seem like the approach to take. 
```
pkg/remotewrite/config.go:289    funlen  Function 'parseEnvs' is too long (84 > 80)
pkg/remotewrite/trend.go:117:21  gosec   G115: integer overflow conversion int -> uint16
```

olegbespalov:
@obanby, no worries. We will take care of these linter issues. We'll merge this PR soon. Thanks again for your contribution! :bow: :tada:

joanlopez:
Thank you @obanby for your contribution! 🙇🏻 
Thanks @olegbespalov for driving this to getting it merged! 💟

